PR #28, opened 2021-05-13
🧬 [EXPERIMENTAL] Explore potential API's for the Dialog component

achambers:
This PR is exploring what a potential `<Dialog />` component might look like. I'm trying to mirror the https://headlessui.dev/react/dialog version as much as possible and so, heavily using the documentation and the React/Vue versions as inspiration.

This is a temporary PR until I settle on a good solution at which point I'll reach out for feedback from @GavinJoyce to discuss whether it's something we want to move forward with.

## 👑 Quest Items

- [ ] `<Dialog />` API including `<Overlay />`, `<Title />` and `<Decscription />`
- [ ] Support for css transitions (likely using ember-css-transitions but I have a few niggles I need to work through)
- [ ] Discuss how the `<Transition />` concept in React translates, if at all, to Ember and how best to implement the same functionality
- [ ] Focus management
- [ ] Accessibility
- [ ] Mouse interaction
- [ ] Keyboard interaction
- [ ] Rendering dialog in a "portal" (using `{{#in-element}}`)

(This PR includes an upgrade of Ember to 3.26.x due to experimenting using the `{{focus-trap}}` modifier which needs 3.22)

PR #31, opened 2021-05-20
✨ Introduce `<Dialog />` component

achambers:
# ✨ What Changed and Why

This PR introduces the `<Dialog />` component based on https://headlessui.dev/react/dialog

This PR implements the component by re-implementing the Vue tests ([LINK](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/dialog/dialog.test.ts)). Where I can mirror these tests I do, but not all of them  apply one-to-one with the Vue implementation. Where they do diverge, I stick with Ember idioms and practices to achieve a similar result.

This PR doesn't aim to implement every single aspect of the reference implementation. Just the core functionality to start using it. The things that this PR does implement are:

- Basic show and hide
- Implementing the `Dialog`, `Overlay`, `Title` and `Description` components
- Styling the Dialog and the Overlay
- Rendering in a portal
- Accessibility (focus management, mouse interaction, keyboard interaction, aria labels)

Things that this PR does not implement are:

- Specifying initial focus element
- `static` prop
- Composition with transitions
- Nesting of `<Dialog>`s

Below is a record of the reference implementation component API's and what has been implemented in this PR:

❌ === not implemented

### Props (Vue) / Args (Ember)
| Vue  | Ember |
| ------------- | ------------- |
| `open`  | `@open`  |
| `initialFocus`  | ❌  |
|`as`| `@as`|
|`static`|❌|
|`unmount`|❌|

### Events (Vue) / Actions (Ember)
| Vue  | Ember |
| ------------- | ------------- |
| `close`  | `@onClose`  |

### Slot props (Vue) / Yielded data (Ember)

| Vue  | Ember |
| ------------- | ------------- |
| `open`  | `open`  |

### Props (Vue) / Args (Ember)
| Vue  | Ember |
| ------------- | ------------- |
| `as`  | `@as`  |

### Slot props (Vue) / Yielded data (Ember)

| Vue  | Ember |
| ------------- | ------------- |
| `open`  | `open`  |

### Props (Vue) / Args (Ember)
| Vue  | Ember |
| ------------- | ------------- |
| `as`  | `@as`  |

### Slot props (Vue) / Yielded data (Ember)

| Vue  | Ember |
| ------------- | ------------- |
| `open`  | `open`  |

### Props (Vue) / Args (Ember)
| Vue  | Ember |
| ------------- | ------------- |
| `as`  | `@as`  |

### Slot props (Vue) / Yielded data (Ember)

| Vue  | Ember |
| ------------- | ------------- |
| `open`  | `open`  |

# 😈 Before

Didn't exist

# 😇 After

https://user-images.githubusercontent.com/416724/120331990-37d67400-c2e6-11eb-8cdf-0a5d4ded706d.mov

Closes #28

### DIFF
--- a/addon/components/dialog.hbs
+++ b/addon/components/dialog.hbs
@@ -1,0 +1,43 @@
+{{#if @isOpen}}
+  {{#let (element this.tagName) as |Tag|}}
+    {{#in-element this.$portalRoot insertBefore=null}}
+      <Tag
+        id={{this.guid}}
+
+        role="dialog"
+        aria-modal="true"
+
+        {{focus-trap}}
+        {{click-outside @onClose}}
+        {{this.handleEscapeKey @isOpen @onClose}}
+        {{on "click" this.stopPropagation}}
+
+        ...attributes
+      >
+        {{yield (hash
+          isOpen=@isOpen
+          onClose=@onClose
+          Overlay=(
+            component "dialog/-overlay"
+            guid=this.overlayGuid
+            dialogGuid=this.guid
+            isOpen=@isOpen
+            onClose=@onClose
+          )
+          Title=(
+            component "dialog/-title"
+            guid=this.titleGuid
+            dialogGuid=this.guid
+            isOpen=@isOpen
+          )
+          Description=(
+            component "dialog/-description"
+            guid=this.descriptionGuid
+            dialogGuid=this.guid
+            isOpen=@isOpen
+          )
+        )}}
+      </Tag>
+    {{/in-element}}
+  {{/let}}
+{{/if}}
--- a/addon/components/dialog.js
+++ b/addon/components/dialog.js
@@ -1,0 +1,91 @@
+import Component from '@glimmer/component';
+import { typeOf } from '@ember/utils';
+import { guidFor } from '@ember/object/internals';
+
+import { modifier } from 'ember-modifier';
+
+import { Keys } from 'ember-headlessui/utils/keyboard';
+
+export default class DialogComponent extends Component {
+  DEFAULT_TAG_NAME = 'div';
+
+  guid = `${guidFor(this)}-headlessui-dialog`;
+  $portalRoot = document.body;
+
+  handleEscapeKey = modifier((_element, [isOpen, onClose]) => {
+    let handler = (event) => {
+      if (event.key !== Keys.Escape) return;
+      if (!isOpen) return;
+
+      event.preventDefault();
+      event.stopPropagation();
+
+      onClose();
+    };
+
+    window.addEventListener('keyup', handler);
+    return () => {
+      window.removeEventListener('keyup', handler);
+    };
+  });
+
+  constructor() {
+    super(...arguments);
+
+    let { isOpen, onClose } = this.args;
+
+    if (isOpen === undefined && onClose === undefined) {
+      throw new Error(
+        'You have to provide an `isOpen` and an `onClose` prop to the `Dialog` component.'
+      );
+    }
+
+    if (isOpen === undefined) {
+      throw new Error(
+        'You provided an `onClose` prop to the `Dialog`, but forgot an `isOpen` prop.'
+      );
+    }
+
+    if (onClose === undefined) {
+      throw new Error(
+        'You provided an `isOpen` prop to the `Dialog`, but forgot an `onClose` prop.'
+      );
+    }
+
+    if (typeOf(isOpen) !== 'boolean') {
+      throw new Error(
+        `You provided an \`isOpen\` prop to the \`Dialog\`, but the value is not a boolean. Received: ${isOpen}`
+      );
+    }
+
+    if (typeOf(onClose) !== 'function') {
+      throw new Error(
+        `You provided an \`onClose\` prop to the \`Dialog\`, but the value is not a function. Received: ${onClose}`
+      );
+    }
+  }
+
+  get tagName() {
+    return this.args.as || this.DEFAULT_TAG_NAME;
+  }
+
+  get overlayGuid() {
+    return `${this.guid}-overlay`;
+  }
+
+  get titleGuid() {
+    return `${this.guid}-title`;
+  }
+
+  get descriptionGuid() {
+    return `${this.guid}-description`;
+  }
+
+  get $element() {
+    return document.getElementById(this.guid);
+  }
+
+  stopPropagation(event) {
+    event.stopPropagation();
+  }
+}
--- a/tests/dummy/app/templates/dialog/dialog-modal.hbs
+++ b/tests/dummy/app/templates/dialog/dialog-modal.hbs
@@ -1,0 +1,63 @@
+<div class="flex items-start justify-center w-screen h-full p-12 bg-gray-50">
+  {{#let (hash isOpen=false) as |state|}}
+    <button
+      type="button"
+      class="inline-flex items-center px-4 py-2 text-base font-medium text-white bg-indigo-600 border border-transparent rounded-md shadow-sm hover:bg-indigo-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500"
+      {{on "click" (set state.isOpen true)}}
+    >
+      Open dialog
+    </button>
+
+    <Dialog
+      class="fixed inset-0 z-10 overflow-y-auto"
+      @as="div"
+      @isOpen={{state.isOpen}}
+      @onClose={{set state.isOpen false}}
+      as |dialog|
+    >
+      <div class="flex items-end justify-center min-h-screen px-4 pt-4 pb-20 text-center sm:block sm:p-0">
+        <dialog.Overlay class="fixed inset-0 bg-gray-500 bg-opacity-75 transition-opacity" />
+
+        {{!-- This element is to trick the browser into centering the modal contents. --}}
+        <span class="hidden sm:inline-block sm:align-middle sm:h-screen" aria-hidden="true">
+          &#8203;
+        </span>
+
+        <div class="inline-block px-4 pt-5 pb-4 overflow-hidden text-left align-bottom bg-white rounded-lg shadow-xl transform transition-all sm:my-8 sm:align-middle sm:max-w-sm sm:w-full sm:p-6">
+          <div>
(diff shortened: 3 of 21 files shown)


achambers:
Hey @GavinJoyce, what version of Ember are you wanting to target with this addon? In this PR I used a local modifier (https://github.com/GavinJoyce/ember-headlessui/pull/31/files#diff-12c6dffeaf696cb3f4f7b58300125eacc6085f079361adcd7a3e94c3528d1b40R15-R23) which only works in >=3.25 I believe.

The intention here was that I wanted to avoid bringing another addon in just for this and it's functionality specific to this modal, for now. Figured it could be abstracted at a later date to something more general if need be.

Thoughts?

GavinJoyce:
I think we can target whichever version makes sense for us. +1 on >=3.25

achambers:
> I think we can target whichever version makes sense for us. +1 on >=3.25

Agreed. I’ll update the ember try scenarios to reflect that then as I believe they are the reason this build is broken.

achambers:
@GavinJoyce I'm just going to put this PR on hold for a short while as our current needs required the "Popover". So, I'm going to go ahead and implement that in a separate PR so we can move forward. Just wanted to touch base to make sure we're not treading on each other's toes.

GavinJoyce:
Sounds good, thanks

achambers:
> Could we use `{{#in-element}}` and append to `document.body`? If there are cases where people want to customise the element, we could consider expanding the API at that point. Would that work for your own use?

Oh, use document body as the destination. That did cross my mind, but I dismissed it. Maybe that was prematurely. I'll give that a crack and see how we get on.

roomman:
Just sharing a bit of me own experience using {{#in-element}} for a dialog. 

I needed to pass in a target node so I could get my render tests to work, as document.body wasn't in scope. 

I ended up using a very simple  helper to grab a node if the user passed in @targetElement. If they don't then it defaults to document.body. Eg:

`{{#in-element (query-selector this.parentElementSelector) insertBefore=null}}`

I hope that's helpful and not missing your point 🤞🏻

achambers:
@GavinJoyce What are your thoughts on focus management (trapping focus inside a modal dialog while it is open).

HeadlessUI implement their own (https://github.com/focus-trap/focus-trap/blob/master/index.js) while there is also a lib that provides this functionality (https://github.com/focus-trap/focus-trap/blob/master/index.js). My initial instinct was to use the focus-trap lib. But looking at the Headless UI implementation, I'm wondering whether it makes sense just to re-implement that ourselves. It seems simpler and maybe is better to avoid too many external dependencies. "Famous last words".

Do you have any initial thoughts or preferences on this?

GavinJoyce:
Both seem like fine options to me. If it's pretty simple, reimplementing their simple version here seems good to me. Whatever you prefer is good with me

addon/components/dialog.hbs
```diff
@@ -0,0 +1,42 @@
+{{#if @open}}
```
GavinJoyce, line 1: I wonder if we should deviate from the headless-ui naming for arguments like this? To my mind, `@open` is ambiguous - it could be an action name. Perhaps `@isOpen` is better? It would also be symmetrical with [`@isOn` in `<Switch />`](https://github.com/GavinJoyce/ember-headlessui/blob/7cc969213c28c09bc3ae07ab62b2831d0aa17696/addon/components/switch.hbs#L7). What do you think?

tests/dummy/app/templates/dialog/dialog-modal.hbs
```diff
@@ -0,0 +1,63 @@
...
+                </p>
+              </div>
+            </div>
+          </div>
+
+          <div class="mt-5 sm:mt-6">
+            <button
+              type="button"
+              class="inline-flex justify-center w-full px-4 py-2 text-base font-medium text-white bg-indigo-600 border border-transparent rounded-md shadow-sm hover:bg-indigo-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500 sm:text-sm"
+              {{on "click" (set state.isOpen false)}}
```
GavinJoyce, line 54: perhaps this would be slightly better as `{{on "click" dialog.onClose}}`?

tests/dummy/app/templates/dialog/dialog-modal.hbs
```diff
@@ -0,0 +1,63 @@
...
+    >
+      Open dialog
+    </button>
+
+    <Dialog
+      class="fixed inset-0 z-10 overflow-y-auto"
+      @as="div"
+      @open={{state.isOpen}}
+      @onClose={{set state.isOpen false}}
+      as |d|
```
GavinJoyce, line 16: a very minor thing, perhaps this would be more readable as `as |dialog|`?

GavinJoyce:
Just a few small comments, this is great!

addon/components/dialog.hbs
achambers: Yeh, I tossed this same question around in my head a bit. I settled on aligning with the reference implementation as much as I could. I guess the flipside is that `@onSomething` could be seen as representing actions.

It's something I'm super precious about so I'm happy to change to align with what's already in the codebase if you wish mate.

tests/dummy/app/templates/dialog/dialog-modal.hbs
achambers: Can do

addon/components/dialog.hbs
achambers: And to round this off a little bit we can see that the React implementation exposes an `onClose` prop ([LINK](https://headlessui.dev/react/dialog#component-api)) while the Vue implementation exposes a `close` event ([LINK](https://headlessui.dev/vue/dialog#component-api)) so each implementation is, obviously, built with the patterns of the framework and while `@isSomething` isn't necessarily an Ember pattern, it is expressive and so, if we agree that this is the pattern we want to embrace in this repo I'm all for it.

I'll make the change.

tests/dummy/app/templates/dialog/dialog-modal.hbs
achambers: The `<Dialog/>` component doesn't yield the `onClose` action. Would you like it too?

I'm kind of figuring that what you pointed out here is sort of down to the simplistic implementation of the dummy app - I was just trying to keep the logic in the template. A real world implementation might have a `closeModal` action on the component or whatever, that could feasibly used when passed in to the Dialog and on any buttons on the modal.

I'm not overly precious about this - it's just the dummy app.

If you think the `<Dialog />` component should expose the `@onClose` action though, I can do that. Just can't, yet, see the benefit myself. But I'm probably missing something :)

achambers: Done 👍

tests/dummy/app/templates/dialog/dialog-modal.hbs
achambers: Done 👍

addon/components/dialog.hbs
achambers: Done 👍

tests/dummy/app/templates/dialog/dialog-modal.hbs
GavinJoyce: > I was just trying to keep the logic in the template

I think that's a good reason to yield the `onClose` action, it allows for the action to be defined in one place. My default is to use template only components, I'm sure that I'd use this myself.

achambers: Okie doke. I'll make the change.

addon/components/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+
+import { modifier } from 'ember-modifier';
+
+import { Keys } from 'ember-headlessui/utils/keyboard';
+
+export default class DialogComponent extends Component {
+  DEFAULT_TAG_NAME = 'div';
+
+  guid = `${guidFor(this)}-headlessui-dialog`;
+  $portalRoot = document.body;
```
alexlafroscia, line 13: Another option would be to target the `.ember-application` element; for the app running normally, this will be the body anyway, but for testing it's down inside the preview window.

tests/dummy/app/templates/dialog/dialog-modal.hbs
achambers: Done 👍

alexlafroscia (commented):
I hope you don't mind a drive-by comment; I just wanted to say I'm _super_ excited to see this PR! I came to the repo looking for this component specifically. Great work!

GavinJoyce (approved):
💥 nice work!

addon/components/dialog.hbs
```diff
@@ -0,0 +1,43 @@
+{{#if @isOpen}}
+  {{#let (element this.tagName) as |Tag|}}
+    {{#in-element this.$portalRoot insertBefore=null}}
+      <Tag
+        id={{this.guid}}
+
+        role="dialog"
+        aria-modal="true"
+
+        {{focus-trap}}
```
alexlafroscia, line 10: If you _wanted_ to support the `initialFocus` argument that's mentioned in the PR description, it could be passed down into the underlying `focus-trap` library's options https://github.com/josemarluedke/ember-focus-trap#with-focus-trap-options

achambers: Good shout. I do plan to implement that. Just wasn’t a priority in this PR. Will keep that in mind. Thanks.

addon/components/dialog.js
achambers: Nice idea.

achambers:
@GavinJoyce I've merged this. Are you able to publish to npm this morning by any chance? We've got a PR that's waiting to use this component and currently pointing to the GH branch.

GavinJoyce:
Sorry for the delay, `0.2.0` has been published

I'll add you to npm so that you can publish yourself in future this evening

achambers:
> Sorry for the delay, `0.2.0` has been published
> 
> I'll add you to npm this evening so that you can publish yourself in future

No probs man. And thanks.

GavinJoyce:
You're an owner now: https://www.npmjs.com/package/ember-headlessui

PR #35, opened 2021-05-22
close the menu on click outside

GavinJoyce:
The menu now closes when clicking outside:

Kapture 2021-05-22 at 16 01 21

### DIFF
--- a/addon/components/menu.hbs
+++ b/addon/components/menu.hbs
@@ -1,6 +1,7 @@
 <div
   id={{this.guid}}
   {{on 'pointerdown' this.onPointerDown}}
+  {{click-outside this.close}}
   ...attributes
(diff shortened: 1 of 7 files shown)


addon/components/menu.hbs
```diff
@@ -1,6 +1,7 @@
 <div
   id={{this.guid}}
   {{on 'pointerdown' this.onPointerDown}}
+  {{click-outside this.close}}
```
GavinJoyce, line 4: @achambers fyi, I also needed your modifier from https://github.com/GavinJoyce/ember-headlessui/pull/31 here

achambers: Coolio. I'll change mine to use that.

PR #38, opened 2021-06-07
Inject Dialog element into app root

alexlafroscia:
When checking out the `Dialog` element, I noticed an issue in the testing environment: you can't actually see the dialog.

This happens because the `Dialog` uses `document.body` as a target for the `in-element` helper. This works fine when the app is running normally, but in the testing environment places the `Dialog` outside of the testing window.

The result is that you can't visually _see_ the `Dialog`, in particular when using the `Dev Mode` feature of QUnit.

CleanShot 2021-06-07 at 12 39 43@2x

By looking up the `rootElement` configuration and using that instead, where available, we can place the `Dialog` element _inside_ the testing root, rather than adjacent to it, so the element is visible again

CleanShot 2021-06-07 at 12 39 03@2x

I pulled this off using `@embroider/macros` to look up the app's config option and make it available to the addon. There are other ways to do this, like `ember-get-config`, but from what I can tell using `@embroider/macros` is the recommended approach in a post-Embroider world.

### DIFF
--- a/addon/components/dialog.js
+++ b/addon/components/dialog.js
@@ -1,16 +1,24 @@
 import Component from '@glimmer/component';
 import { typeOf } from '@ember/utils';
 import { guidFor } from '@ember/object/internals';
+import { getOwnConfig } from '@embroider/macros';
 
 import { modifier } from 'ember-modifier';
 
 import { Keys } from 'ember-headlessui/utils/keyboard';
 
+function getPortalRoot() {
+  const { rootElement } = getOwnConfig();
+
+  // If we looked up a `rootElement` config at build-time, use that; otherwise use the body
+  return rootElement ? document.querySelector(rootElement) : document.body;
+}
+
 export default class DialogComponent extends Component {
   DEFAULT_TAG_NAME = 'div';
 
   guid = `${guidFor(this)}-headlessui-dialog`;
-  $portalRoot = document.body;
+  $portalRoot = getPortalRoot();
 
   handleEscapeKey = modifier((_element, [isOpen, onClose]) => {
     let handler = (event) => {
--- a/index.js
+++ b/index.js
@@ -2,4 +2,16 @@
 
 module.exports = {
   name: require('./package').name,
+
+  options: {
+    '@embroider/macros': {
+      setOwnConfig: {},
+    },
+  },
+
+  config(_emberEnv, config) {
+    // Pass `rootElement` config into a location where it can be looked up at run-time
+    this.options['@embroider/macros'].setOwnConfig.rootElement =
+      config.APP.rootElement;
+  },
 };
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "@ember/render-modifiers": "^1.0.2",
+    "@embroider/macros": "^0.41.0",
     "ember-cli-babel": "^7.26.3",
     "ember-cli-htmlbars": "^5.7.1",
     "ember-element-helper": "^0.3.1",


index.js
```diff
@@ -2,4 +2,16 @@
...
+    '@embroider/macros': {
+      setOwnConfig: {},
+    },
+  },
+
+  config(_emberEnv, config) {
+    // Pass `rootElement` config into a location where it can be looked up at run-time
+    this.options['@embroider/macros'].setOwnConfig.rootElement =
+      config.APP.rootElement;
+  },
```
alexlafroscia, line 16: This approach was more-or-less borrowed from the `@embroider/macros` docs

https://github.com/embroider-build/embroider/tree/master/packages/macros#getownconfig-getconfig-and-getglobalconfig

But tweaked slightly to set the config from a location where we know what the consume app's `config/environment.js` file produced

alexlafroscia:
I'm not sure why CI didn't run for this; I tried to tweak the config in #39 but it didn't run CI on the PR with that config, either.

alexlafroscia:
🤦 welp, that explains it!

CleanShot 2021-06-07 at 13 23 28@2x

GavinJoyce:
CI is running now

achambers:
Weird, I thought I did something similar. Apparently I dreamed that ;) Thanks @alexlafroscia

alexlafroscia:
@achambers I think it was part of your plan and was just missed before getting it over the finish line

https://github.com/GavinJoyce/ember-headlessui/pull/31/files#r643253076

No problem!

achambers:
> @achambers I think it was part of your plan and was just missed before getting it over the finish line
> 
> https://github.com/GavinJoyce/ember-headlessui/pull/31/files#r643245957
> 
> This ends up being even more flexible than targeting `.ember-application` specifically, as we're no longer tied to that implementation detail, but in practice the `rootElement` _is_ the `.ember-application` either way!

I definitely wrote the code that uses rootElement from config. Just not sure where it ended up 🤔🤣

GavinJoyce:
🍻

addon/components/dialog.js
```diff
@@ -1,16 +1,24 @@
...
 
 import { modifier } from 'ember-modifier';
 
 import { Keys } from 'ember-headlessui/utils/keyboard';
 
+function getPortalRoot() {
+  const { rootElement } = getOwnConfig();
+
+  // If we looked up a `rootElement` config at build-time, use that; otherwise use the body
+  return rootElement ? document.querySelector(rootElement) : document.body;
```
achambers, line 14: @alexlafroscia hrm.....in our app we're seeing that this doesn't seem to be working. Just starting out trying to dive in and discover why that is, but thought I'd reach out and see if you'd noticed this not working as expected. I don't have any prior experience with the embroider macros so I'm a little out of my depth at the moment.

For clarification, we're seeing that in tests `rootElement` is undefined so the dialog is always injected in to `document.body`. Have you actually see this working in tests where it injects the dialog in to the test container instead of `document.body`?

alexlafroscia: Yeah, this seemed to work from what I could tell... in the tests for this addon itself, this change made the dialog actually visible when testing using "dev mode". Before this PR, the dialog would end up behind the "dev mode" window and would thus be invisible...

achambers: Yeh, I'm wondering if there might be a subtle difference between the tests in this repo vs the tests in a consuming app.

If you could verify it with your app that'd be ace. I'll keep trying to dig in from my end to try and get some more info.

achambers: @alexlafroscia Gaaaaah!!!! You know what I think it is. I'm running the tests by going to https://localhost:4200/tests after running `ember s`. This is as opposed to running `ember t -s`.

For the most part this is ok, but there are subtle differences. And one of them is hitting me right now. The former runs with the EmberCLI `development` environment. The latter runs with the `test` environment. `rootElement` is defined only for the `test` environment. So, I imagine if I running this with `ember t -s` I'll see that it works.

Going to verify now. Apologies for a bum steer if so.

alexlafroscia: Ah, yup, that makes sense!

I run into stuff like this with my team every so often; it's so tempting to use that `/tests` URL, but I always advise against it for reasons like this. _Usually_ the behavior in those environments is the same, but not always!

PR #39, opened 2021-06-07
chore: simplify `on` config for GitHub Actions

alexlafroscia:
I noticed that #38 didn't run CI for some reason, which I've never seen before. I'm wondering if it's because of the specific branch filtering, so I'm trying to remove that to see if this will run CI for a fork PR instead.


alexlafroscia:
Hmm... Nope, doesn't seem like that did it either...

PR #40, opened 2021-06-07
chore: upgrade `ember-truth-helpers` to v3

alexlafroscia:
I noticed the old version when introducing `ember-headlessui` into my app, as the `2.X` version installed here conflicted with the `3.X` version that I was already using.

Since the `3.X` release has been out for a while and the Node/Ember version requirements seem compatible, this _should_ be a non-breaking change to consuming apps.

I see this upgrade is also in #29 but since that PR hasn't landed, it would be nice to upgrade this one package at least, since it impacts consumers of this addon.


GavinJoyce:
Thanks

Issue #41, opened 2021-06-07
Not Compatible with `ember-click-outside-modifier`


alexlafroscia:
I have [`ember-click-outside-modifier`](https://github.com/lifeart/ember-click-outside-modifier#readme) installed in my app. It seems as though `ember-headlessui` breaks it, due to both packages introducing a `click-outside` modifier that are implemented slightly differently.

`ember-click-outside-modifier` passes the event object along as an argument to the action that it calls, while the modifier provided by this addon does not. This addon's modifier is taking precedence (maybe due to it coming alphabetically after the other addon?), so actions in my app that expect to receive an event argument no longer do.

I see a few decent options available, any of which I'm happy to open a PR for; I just wanted to get some feedback on what would be most desirable first!

* Ditch the local modifier and add `ember-click-outside-modifier` as a dependency. This is a nice win for people already using both, and allows each package to do their respective jobs without overlap.
* Avoid registering the `click-outside` modifier that ships with this addon to the global registry at all, instead opting to make it "local" to the components where it's used. This introduces duplication for folks that have `ember-click-outside-modifier` installed, but ensures that they don't compete
* Forward the `event` argument along to the action so that the API for each modifier is the same. Then this helper could be used and `ember-click-outside-modifier` uninstalled
* Configure `ember-headlessui` to be run "before" `ember-click-outside-modifier` so the slightly-more-featureful version of the modifier is actually used in the app.

In the meantime, in case anyone else runs into this issue, I'm going to resolve it with a [`yarn` patch](https://yarnpkg.com/cli/patch/#gatsby-focus-wrapper) to apply the "before" config to my local installation of the addon.

alexlafroscia:
No longer relevant, as we now _use_ `ember-click-outside-modifier` 😬

PR #43, opened 2021-06-08
Replace `click-outside` modifier with addon

alexlafroscia:
Offering this PR as a solution to #41, though I'm happy to take one of the other approaches I mentioned in there instead!

I chose this once since it's nice to have one tool for a job, across the community; if there's already an addon specifically for this, taking care of bug fixing and improvements, then it makes sense to lean on that rather than having a separate tool covering the same concern.


alexlafroscia:
So, the test failure here is interesting!

Essentially, what's happening is this sequence of events:

1. Click on the trigger button to launch open modal
2. Modal opens, installing a `click` handler to an element above the trigger button _while the event is still propagating_
3. Parent element catches the `click` event from the button having been pressed
4. Modal closes again immediately

This seems undesirable from the perspective of the user, but also not _wrong_ per say. I was confused at first about why this is happening, but realizing that we install a handler for an event while the event is still being fired made me realize that what's happening does make a certain kind of sense.

I think it wouldn't be unreasonable for the trigger button to need to call `stopPropagation` to prevent this externally, though that's certainly a "gotcha" worth considering documentation for. The fact that this happens isn't really a bug in `ember-headlessui` or `ember-click-outside-modifier`; it's just how the DOM works.

On the other hand, since the original `mouseup` binding does _not_ have this problem, while `click` does, maybe it makes more sense to rename the modifier something besides `click-outside` rather than using the addon, or making the modifier "private" to avoid registering it to the global namespace which introduces the conflict with `ember-click-outside-modifier`.

alexlafroscia:
Huh, so adding the `useCapture` param on `click-outside` fixed it 😀

alexlafroscia:
A combination of two changes makes this work again

1. Modify `ember-click-outside-modifer` to bind to the `window` object rather than the `html` element (I'm not sure which is more correct or why to listen on one or the other)
2. Set the `useCapture` flag when listening for the event (which is part of the `ember-click-outside-modifier` API

alexlafroscia:
I think the safest thing would be to not change the event used to in the `click-outside` modifier, so I opened a PR to _that_ addon that would allow us to continue binding to the same event as before! Best of both worlds (no duplication, same behavior as what's working right now)

GavinJoyce:
Thanks!

achambers:
Sorry for the late contribution here, just swinging back around to this.....My $0.02 is that there is something nice, and reassuring about this addon being pretty self contained and relying, as little as possible, on outside dependencies. The surface are of this addon is pretty well defined and constrained, something like the "click outside" functionality is also quite constrained and an implementation likely isn't going to change too much over time. So, I think there is something to be said for having this in our own code. The same goes for the focus-trap functionality.

Not saying that we should go that path for either of those areas, but, I do think this is possibly something worth considering as we move forward - do we value minimizing dependencies so that this addon can be well constrained and not at the mercy of other repos?

alexlafroscia:
That's a great point to raise @achambers!

From the perspective of this change, I think the alternate solution that could have been taken where the `click-outside` became a "local" modifier, like some of the other ones that are used around this repo, would have both removed the conflict with `ember-click-outside-modifier` _and_ avoided the external dependency.

The trade-off becomes, for apps that are also using `ember-click-outside-modifier`, having to ship some duplicated code that could instead rely on a dependency they already have. It's a small amount of code, though, so maybe that doesn't really matter!

I wouldn't mind if this addon took a stance of implementing stuff like that itself, as long as it also avoids registering it's own versions of those modifiers into the global namespace so that it avoids potential conflicts with other packages.

achambers:
> That's a great point to raise @achambers!
> 
> From the perspective of this change, I think the alternate solution that could have been taken where the `click-outside` became a "local" modifier, like some of the other ones that are used around this repo, would have both removed the conflict with `ember-click-outside-modifier` _and_ avoided the external dependency.
> 
> The trade-off becomes, for apps that are also using `ember-click-outside-modifier`, having to ship some duplicated code that could instead rely on a dependency they already have. It's a small amount of code, though, so maybe that doesn't really matter!
> 
> I wouldn't mind if this addon took a stance of implementing stuff like that itself, as long as it also avoids registering it's own versions of those modifiers/helpers/whatever into the global namespace so that it avoids potential conflicts with other packages.

@alexlafroscia My preference would also be that this addon remained as self contained as possible too. 

For context, the outside click modifier started out as a local modifier on the dialog component. It was then needed for the Menu component and subsequently moved to the global namespace. We didn’t consider the ramifications of it clashing with other addons to be honest. 

My preference here would be to have the logic encapsulated in some sort of utility function that can be improved at the various call sites that need to expose it as a local modifier.

Issue #42, opened 2021-06-08
Auto-Deploy Docs Site to GitHub Pages


alexlafroscia:
The docs site for `ember-headlessui` is really nice! It would be great to make it easier for people checking out this addon to see it.

Would you be interested in having it automatically deploy to GitHub Pages? I do that for a number of my addons, like `ember-steps`, and the configuration for it is pretty easy! I would be happy to set that up if you're interested.

GavinJoyce:
That would be great, thanks for offering. LMK if you need access to anything to set it up

alexlafroscia:
I _might_ need a Github "Personal Access Token" configured in GitHub Pages, but I think that they now allow Actions to upload to GitHub Pages without that. If I do, then that's something you could generate and configure in the repo settings. I'll let you know!

alexlafroscia:
Seems like the deploy worked! It's on the `gh-pages` branch, at least

https://github.com/GavinJoyce/ember-headlessui/tree/gh-pages

It doesn't see like the URL is resolving yet, but I _think_ this should be it when it's done?

http://gavinjoyce.github.io/ember-headlessui/

Also, I realize I named the actual CI jobs badly. I'll clean that up at some point 😬

alexlafroscia:
Site still isn't live, but I realized I can validate what might be happening on my fork! 🤦 

I had to go to this URL (modified to match what it should be for this repo)

https://github.com/GavinJoyce/ember-headlessui/settings/pages

and select the `ph-pages` branch from the drop-down menu. Once I did that, the site (for my branch) went live! 🎉 

https://alexlafroscia.github.io/ember-headlessui/

(I will go and disable that again once the "real" repo has the docs site turned on!)

alexlafroscia:
While the initial page looks correct, the links on the main page don't actually go to the right location -- they aren't respecting the root URL configuration. Lemme work out how to get it fixed on my branch and I'll open a PR to fix the upstream repo before you go and turn on the docs site here!

GavinJoyce:
The https://gavinjoyce.github.io/ember-headlessui/ index page is loading now

alexlafroscia:
Great!

It turns out the issue is that the links on the site use normal anchors instead of `LinkTo`. I'll have a PR open in a moment to fix that!

alexlafroscia:
Sweet, looks like the site's all working now!

PR #44, opened 2021-06-08
Deploy Docs Site to GitHub Pages

alexlafroscia:
First attempt at #42 

This might be a trial-and-error kind of situation if this doesn't work right off the bat, but I'm pretty sure this should be fine. I know that we can use `ember-cli-deploy` or something in here, as I do with `ember-steps` for example, but I honestly don't think it's necessary. Setting the right `rootURL` should be all we need to do, since this action handles actually putting the assets on the branch!

https://github.com/marketplace/actions/deploy-to-github-pages

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,33 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


GavinJoyce:
Nice. Is this ready to merge and try out?

alexlafroscia:
@GavinJoyce I think so! We can't really debug much here until the workflow runs, so I'll take a look at the log for the job once it's running and open a PR with fixes if needed.

GavinJoyce:
🤞

PR #46, opened 2021-06-10
fix(switch): default to using a `button` element

alexlafroscia:
I've been checking out the `Switch` component and noticed a difference between the behavior here and in the "real" version of HeadlessUI. In the React and Vue components, the actual switch itself defaults to a `button` but can be overwritten to something else. Here, the default is a `div`!

The documentation on the Switch is a little hard to compare 1:1 to the Ember version, but if you look here

https://headlessui.dev/react/switch#switch

it mentions that the default tag is a `button`. Looking at the actual example of a switch in the docs site, it's indeed a `button` under-the-hood!


GavinJoyce:
👍 thanks

PR #47, opened 2021-06-11
Set initial focus for dialog

far-fetched:


tests/accessibility-assertions.js
```diff
@@ -38,7 +38,7 @@ function assertActiveElement(element) {
       return;
     }
 
-    Qunit.assert.equal(element, document.activeElement);
+    Qunit.assert.equal(document.activeElement, element);
```
far-fetched, line 41: It looks bizarre in case of error, as the second argument is 'expected'  and the first 'actual'.

achambers: What do you mean in the case of error? You mean when this assertion fails? Can you give me an example.

Maybe I'm missing something because `Qunit.assert.equal(element, document.activeElement);` reads correctly to me in that I want to assert that this particular element (actual) is indeed the active element (which we know is the document.activeElement - expected).

That seems accurate to me. What am I missing @far-fetched ?

tests/integration/components/dialog-test.js
```diff
@@ -499,4 +499,29 @@ module('Integration | Component | <Dialog>', function (hooks) {
...
+        <Dialog
+          @isOpen={{this.isOpen}}
+          @onClose={{set this.isOpen false}}
+          @initialFocus={{this.button}}
+        >
+          <input type="text" value="first">
+          <button type="button">focused</button>
+        </Dialog>
+      `);
+      this.set('button', buttonElement);
```
achambers, line 521: Oh, I see....you're using the fact that focus-trap `initialFocus` can accept a function that returns an HTMLElement. I wonder whether for simplicity of the test, we should just us a css selector, maybe something like this:

```hbs
<button type="button" {{on "click" (set this.isOpen true)}}>
  Trigger
</button>
<Dialog
  @isOpen={{this.isOpen}}
  @onClose={{set this.isOpen false}}
  @initialFocus="[data-test-focussed-button]"
>
  <input type="text" value="first">
  <button data-test-focussed-button type="button">focused</button>
</Dialog>
```

That way we can remove the `buttonElement` function altogether.

tests/integration/components/dialog-test.js
```diff
@@ -499,4 +499,29 @@ module('Integration | Component | <Dialog>', function (hooks) {
...
+      await click(getByText('Trigger'));
+      await settled();
```
achambers, line 523: Surprised to see this here. Wasn't needed in any other tests. What's the reason for needing this?

Also, I can't see that it's imported either. How is that working?

achambers (commented):
Nice. Thanks @far-fetched . Have left a couple of questions there.

tests/accessibility-assertions.js
far-fetched: When test fails, with previous version: `Qunit.assert.equal(element, document.activeElement);`:
Captura de pantalla 2021-06-11 a las 10 47 43

But we expect that button will be focused.
So after this change:
 
Captura de pantalla 2021-06-11 a las 10 53 01

For me, element passed to this helper is not actual (we rather expect that passed element is actual) - Actual focused element is - `document.activeElement` so it should be first passed to `equal`. Does it make sense to you ?

tests/integration/components/dialog-test.js
far-fetched: Forgot to commit import 😅 
Without it, focus remains on the clicked open dialog button (`trigger` label). With settled is properly moved to expected button on the dialog.
Actually now I see that I got this error: 
https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-settled-after-test-helper.md
maybe we are with mentioned issue ? 
`In some cases this pattern is mistakenly used to "wait a little more", which usually indicated a deeper root cause issue, like a race condition or missing test waiter somewhere. This pattern only works sometimes because it causes the test to continue on the next tick of the JS runloop, instead of continuing directly. It is highly recommended to fix the root cause in these cases instead of relying on such brittle workarounds.`
Or any different ideas ?

tests/accessibility-assertions.js
achambers: 👍

tests/integration/components/dialog-test.js
achambers: Yeh, so it feels like we _might_ need to register a test waiter potentially. I'd need to dive a little deeper to understand what's going on. Fancy pairing on this to get to the bottom of it?

alexlafroscia: Another approach for handling asserting against conditions that are going to _eventually_ be true is to use [`qunit-wait-for`](https://github.com/alexlafroscia/qunit-wait-for). It allows you to repeat an assertion until it passes or an eventual timeout occurs.

In this case, it would look something like removing `settled` and wrapping the following assertion like this

```js
await assert.waitFor(() => {
  assertActiveElement(getByText('focused'));
});
```

This allows us to avoid needing to instrument the component for testing purposes but still makes the tests resilient to a delay between the trigger being clicked and element becoming focused.

achambers: Thanks for the suggestion @alexlafroscia . I tend to feel like `waitFor` is a little smelly, but can't really articulate why. But, in this situation, it's probably a sensible solution. We discussed some other options but all of them would have involved making changes to the component design just to accommodate the test, which isn't ideal.

So yeh, @far-fetched, want to try `waitFor`? It seems like a reasonable middle ground :) 

Thanks @alexlafroscia  :)

alexlafroscia: It can definitely feel a little weird, in a "doing something in the test to make the test pass instead fo making the code work how I expected" kind of way! It's a pretty common pattern in other UI-testing frameworks, though; QUnit just doesn't ship with something for this out-of-the-box. Even Ember's testing helpers have `waitUntil` which work the same way (and truthfully, from which I borrowed a good amount of the `waitFor` code 😉) -- this just couples the waiting with the assertion!

far-fetched: Seems pretty elegant now !

achambers (approved):
Legendary 🙌 . Thanks @far-fetched

achambers:
Released as https://github.com/GavinJoyce/ember-headlessui/releases/tag/v0.4.0

PR #50, opened 2021-06-11
docs(switch): add example of building a checkbox

alexlafroscia:
I realized after checking out the `Switch` component that it's a great candidate for the foundation of a custom-styled checkbox! Since you can't style an `input[type="checkbox"]` natively and have to build something custom, building on top of `Switch` helped me get the accessibility requirements correct.

All of the `aria` behavior for a switch and a checkbox are the same, save the `role="checkbox"` on the button instead of the default `role="switch"`. That's easy to override in the implementation, though, since you can override default attributes!

I figured this might be a useful "cookbook" example for people since the Headless UI example doesn't show usage as a checkbox, and it's worth demonstrating how `switch.Button` can/should override the `role` attribute.

Here are the two states

CleanShot 2021-06-11 at 11 16 05@2x

CleanShot 2021-06-11 at 11 16 06@2x

### DIFF
--- a/tests/dummy/app/router.js
+++ b/tests/dummy/app/router.js
@@ -16,6 +16,7 @@ Router.map(function () {
 
   this.route('switch', function () {
     this.route('switch-basic');
+    this.route('switch-checkbox');
   });
 
   this.route('dialog', function () {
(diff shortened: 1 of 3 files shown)


achambers (approved):
Love yer work @alexlafroscia 🙌

alexlafroscia:
Thank you @achambers! 😀

PR #53, opened 2021-06-15
Fix `<Menu/>` check if button exists

### LINKED ISSUE #52: 🐛  `<Menu>` focus button after close items
Clicking outside the menu items should close menu, and focus button.

[Test](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/menu-test.js#L539)

[Code](https://github.com/GavinJoyce/ember-headlessui/blob/master/addon/components/menu.js#L36)

Bug:

### **Steps to reproduce the problem**

* menu is rendered on the dialog, 
* clicking outside dialog (close dialog and menu)

### **What went wrong?**

* Button doesn't exist and cannot be focused.

Captura de pantalla 2021-06-15 a las 9 52 05

https://user-images.githubusercontent.com/11621383/122016204-1e8ef680-cdc1-11eb-979e-238beedb3c8a.mov

### **What is the expected behavior?**

Does not focus anything ?

far-fetched:
Fixes #52

### DIFF
--- a/addon/components/menu.js
+++ b/addon/components/menu.js
@@ -33,7 +33,7 @@ export default class Menu extends Component {
     if (this.isOpen) {
       this.isOpen = false;
       next(() => {
-        document.getElementById(this.buttonGuid).focus();
+        this.buttonElement && this.buttonElement.focus();
       });
     }
   }
--- a/tests/integration/components/menu-test.js
+++ b/tests/integration/components/menu-test.js
@@ -564,6 +564,30 @@ module('Integration | Component | <Menu>', (hooks) => {
 
         assert.dom('[data-test-menu-button]').isFocused();
       });
+
+      test('it should not focus button when does not exist', async function (assert) {
+        this.set('isShowButton', true);
+        await render(hbs`
+          <Menu data-test-menu as |menu|>
+            {{#if this.isShowButton}}
+              <menu.Button data-test-menu-button>Trigger</menu.Button>
+            {{/if}}
+            <menu.Items data-test-menu-items as |items|>
+              <items.Item as |item|>
+                <item.Element>Item A</item.Element>
+              </items.Item>
+            </menu.Items>
+          </Menu>
+        `);
+
+        await click('[data-test-menu-button]');
+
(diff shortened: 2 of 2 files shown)


GavinJoyce:
Thanks for this @far-fetched. What do you think about adding a test?

far-fetched:
I wasn't sure that brings any value. But if you think that, I can do it.

addon/components/menu.js
```diff
@@ -33,7 +33,7 @@ export default class Menu extends Component {
     if (this.isOpen) {
       this.isOpen = false;
       next(() => {
-        document.getElementById(this.buttonGuid).focus();
+        document.getElementById(this.buttonGuid)?.focus();
```
achambers, line 36: I'm always a little confused about Babel, but does the optional chaining used here require the consuming app to have the Babel plugin (https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining.html) installed?

Is there any chance this could cause grief to app's who's users don't use a browser that yet supports optional chaining (https://caniuse.com/mdn-javascript_operators_optional_chaining) ?

Should we just go the long hard form here to be safe?

/cc @far-fetched @GavinJoyce

far-fetched: I think it will break with old versions. I guess the question is how far we want to go with browser support.

achambers: I think where this case is concerned, my preference is to go with old school `if`. Can’t see a lot to be gained using optional chaining here.

achambers (approved):
Thanks @far-fetched

achambers:
Actually, before I merge this @far-fetched , will you please rebase and squash the commits together. Both 9415928a61fc8c4a0542d08e9b6d73365b245edb and 9415928a61fc8c4a0542d08e9b6d73365b245edb are unnecessary in "telling the story" of this change.

far-fetched:
🚀

achambers:
Thanks @far-fetched . Released as https://github.com/GavinJoyce/ember-headlessui/releases/tag/v0.4.1

PR #56, opened 2021-06-26
Format `.hbs` files with Prettier

alexlafroscia:
Prettier is great for lots of reasons, primarily because it takes the guess-work out of how to correctly format files. This is great for OSS in particular, since it helps contributors make changes with confidence, knowing that formatting will not be an issue in a PR review.

This PR adds Prettier as a rule for `ember-template-lint` and runs the formatter on all of the files.

**Pros:**

* Easier to make contributions when formatting is handled by the tooling
* Files are all consistent now

**Cons:**

* Some components had line breaks between specific groups of attributes, likely because they are related for accessibility purposes. Those line breaks are now gone.


tests/dummy/app/templates/switch/switch-checkbox.hbs
```diff
@@ -1,22 +1,22 @@
...
+      role='checkbox'
     >
-      <div class="
-        h-full w-full transition duration-200
-        {{if switch.isOn "bg-indigo-600" "bg-gray-200"}}
-      " />
+      <div
+        class='h-full w-full transition duration-200
+          {{if switch.isOn 'bg-indigo-600' 'bg-gray-200'}}
+          '
```
GavinJoyce, line 17: prettier didn't do a great job here IMO, but it's probably an acceptable trade off for the benefits that it brings

GavinJoyce:
Having worked a little on hbs for prettier, I'm surprised that it did such a reasonable job.

There are a few places where the formatted output is worse than the input, but I think this is an acceptable trade-off

GavinJoyce:
Thanks for this @alexlafroscia 🍻

alexlafroscia:
You're welcome! The Glimmer support in Prettier has gotten... "okay" recently, enough that I tend to lean more toward using it than not using it. I agree that the formatting of concat statements (the mustache-embedded-in-a-string syntax) can be pretty weird, but I'd personally take consistent and weird over pretty-but-needs-to-be-managed-manually!

PR #57, opened 2021-06-26
feat(menu): implement search functionality

alexlafroscia:
This implements the currently-missing ability to activate menu items by pressing the first (or first few) characters of the menu item's text.

**Pros:**

- Implements a missing feature that the React and Vue versions have

**Cons:**

- There is a _slight_ delay when running the tests for this functionality, due to the `timeout(350)` that clears the search term. This could be resolved by using a different timeout length in a test build, but I don't like to introduce differences in behavior between tests and the "real" implementation unless it's _really_ necessary, and it just doesn't feel that necessary to me in this case.
- Introduces an additional dependency on `ember-concurrency`, and I know that @achambers is (rightfully) hesitant to add additional dependencies. I believe this one is justifiable for a handful of reasons, but am happy to re-build this _without_ it if these reasons aren't enough:
  - The host app probably already depends on `ember-concurrency` due to how common (and useful) it is
  - There's little risk of `ember-concurrency` going un-maintained in the future due to it's prevalence in the community
  - Getting the logic around cancellation of events and timeouts can be _really_ hard; this is exactly what `ember-concurrency` is great at! If you look at the logic for the search functionality in the "official" implementations, it is _way_ more complicated than the task-based implementation in this PR.

### DIFF
--- a/tests/integration/components/menu-test.js
+++ b/tests/integration/components/menu-test.js
@@ -64,6 +64,20 @@ function assertNoActiveMenuItem(menuSelector) {
   QUnit.assert.dom(menuSelector).doesNotHaveAria('activedescendant');
 }
 
+async function type(selector, value) {
+  let finalKeyEventProcessing;
+
+  for (const char of value) {
+    finalKeyEventProcessing = triggerKeyEvent(
+      selector,
+      'keydown',
+      char.toUpperCase()
+    );
+  }
+
+  await finalKeyEventProcessing;
+}
+
 module('Integration | Component | <Menu>', (hooks) => {
   setupRenderingTest(hooks);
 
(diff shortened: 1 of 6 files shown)


GavinJoyce:
Thanks for this @alexlafroscia. 

👍 on the inclusion of `ember-concurrency`, I haven't worked on any non-trivial app that didn't use it.

If it hasn't been reviewed by then, I'll give this PR a proper review later today

GavinJoyce:
This is great, thanks!

tests/integration/components/menu-test.js
```diff
@@ -64,6 +64,20 @@ function assertNoActiveMenuItem(menuSelector) {
...
+    finalKeyEventProcessing = triggerKeyEvent(
+      selector,
+      'keydown',
+      char.toUpperCase()
+    );
+  }
+
+  await finalKeyEventProcessing;
+}
+
```
alexlafroscia, line 80: Just wanted to throw an explanation in here, since this is a little weird:

We need a custom way to trigger a sequence of key events because Ember's `fillIn` and `typeIn` helpers reject sending input to an element that isn't an `input` or `contenteditable`.

We only want to await the _final_ key event, otherwise we will wait the 350ms after _each_ event for the search term to be cleared, which is _not_ what we want.

I tried to implement the Ember Concurrency task using `rawTimeout` (which the tests will _not_ wait for) but then ran into a problem where it's quite for the test to wait for the search term to clear when we _do_ want it to, since in some tests we search for multiple items in a row. 

TL;DR: I tried to clean this up but the other approaches I could think of didn't actually work in practice!

achambers:
> 👍 on the inclusion of `ember-concurrency`, I haven't worked on any non-trivial app that didn't use it.

💯 has my support too

PR #58, opened 2021-06-26
`Menu` Component Improvements

### LINKED ISSUE #54: `<Menu>` doesn't close items dropdown on `Escape`
We should make the `<Menu>` dropdown close when Escape is hit. I suspect it should go here.

https://github.com/GavinJoyce/ember-headlessui/blob/57f3da16050f1eeb608d9a2bbea993c75e24c857/addon/components/menu/items.js#L6-L20

### LINKED ISSUE #55: Focus of menu button on close seems incorrect.
When I click out side of the menu, say on another button, when the menu closes it sets focus back on the menu button whereas I'd expect the focus to remain on the element I just clicked on. I'm not clear if it makes sense to refocus the button at all. Maybe it does under some circumstances but unsure what they are. In short though, the example below doesn't make sense so there is some improvement that could be made:

In this screencast I'm doing the following:

1) Click on the 3rd dropdown
2) Click on the 2nd dropdown
3) Press up/down keys which opens the 3rd dropdown again.

https://user-images.githubusercontent.com/416724/122551418-dd028380-d02c-11eb-9529-f8df5b587fa5.mov

I think the offending code is here:
https://github.com/GavinJoyce/ember-headlessui/blob/57f3da16050f1eeb608d9a2bbea993c75e24c857/addon/components/menu.js#L35-L37

alexlafroscia:
## `Menu` Avoids Rendering a DOM Node

The `Menu` component in the React/Vue implementations does not render a DOM node by default. It _can_, because I believe they want to make sure that all components can use an `as=` property to configure the node that's rendered, but by default it just renders a `Fragment` (which results in nothing in the DOM).

Since Ember doesn't really have a concept of a `Fragment`, I think that not rendering a DOM node at all is closer to the React/Vue implementation than rendering a `div`.

**Note:** This is a breaking change, as any attributes or modifiers placed on this element should be moved to a new parent `div` element

### Before

```hbs
<Menu data-test-menu class="bg-gray" as |menu|>
  ...
</Menu>
```

### After

```hbs
<div data-test-menu class="bg-gray">
  <Menu as |menu|>
    ...
  </Menu>
</div>
```

## `Menu::Items` Focusing on Render

From the Headless UI documentation on the Menu:

> Clicking the `MenuButton` toggles the menu and focuses the `MenuItems` component.

Using the focus trap here is a little unintuitive, because the menu item elements themselves are not focusable. However, this is a useful approach because

* Focus should be locked within the menu anyway, per the accessibility requirements of a menu
* Focus needs to be returned to the previously-focused element when the menu closes, which the `focus-trap` modifier will do for us

The documentation for the `focus-trap` library describes how you should handle a case where the element that focus is trapped within has no focusable children, which is to focus the "wrapper" element itself using `initialFocus`, which is the approach taken here!

Using the `focus-trap` library to return focus to the trigger button should resolve the weirdness around focus returning to the wrong element sometimes. `focus-trap` will also handle the "press-`escape`-to-dismiss-the-menu" behavior for us without needing additional logic for that.

Closes #55
Closes #54

### DIFF
--- a/addon/components/menu.hbs
+++ b/addon/components/menu.hbs
@@ -1,42 +1,34 @@
-<div
-  id={{this.guid}}
-  ...attributes
-  {{on 'pointerdown' this.onPointerDown}}
-  {{click-outside this.close event='mouseup'}}
->
-  {{yield
-    (hash
+{{yield
+  (hash
+    isOpen=this.isOpen
+    Button=(component
+      'menu/button'
+      buttonGuid=this.buttonGuid
+      itemsGuid=this.itemsGuid
       isOpen=this.isOpen
-      Button=(component
-        'menu/button'
-        guid=this.buttonGuid
-        menuGuid=this.guid
-        itemsGuid=this.itemsGuid
-        isOpen=this.isOpen
-        onClick=this.toggle
-        openMenu=this.open
-        closeMenu=this.close
-        goToFirstItem=this.goToFirstItem
-        goToLastItem=this.goToLastItem
-        goToNextItem=this.goToNextItem
-        goToPreviousItem=this.goToPreviousItem
-      )
-      Items=(component
-        'menu/items'
-        guid=this.itemsGuid
-        buttonGuid=this.buttonGuid
-        isOpen=this.isOpen
-        closeMenu=this.close
-        activeItem=this.activeItem
-        registerItem=this.registerItem
-        unregisterItem=this.unregisterItem
-        goToFirstItem=this.goToFirstItem
-        goToLastItem=this.goToLastItem
-        goToNextItem=this.goToNextItem
-        goToPreviousItem=this.goToPreviousItem
-        goToItem=this.goToItem
-        search=(perform this.searchTask)
-      )
+      openMenu=this.open
+      closeMenu=this.close
+      toggleMenu=this.toggle
+      goToFirstItem=this.goToFirstItem
+      goToLastItem=this.goToLastItem
+      goToNextItem=this.goToNextItem
+      goToPreviousItem=this.goToPreviousItem
     )
-  }}
-</div>
+    Items=(component
+      'menu/items'
+      buttonGuid=this.buttonGuid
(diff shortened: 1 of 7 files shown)


alexlafroscia:
Something isn't quite right with the way the focus trap is being used -- I'll mark this is ready once I fix that up!

GavinJoyce (approved):
I think the existing API is a little nicer to work with, but 👍 on the change if it brings us closer to the React/Vue APIs

alexlafroscia:
> I think the existing API is a little nicer to work with

Just to clarify, you'd prefer that `Menu` _does_ have a DOM node of it's own?

I'm thinking about a case where someone builds out a component that wraps `Menu`. If it were me, I would do something like this:

```hbs
<Menu as |menu|>
  <menu.Trigger
    {{did-insert (set this 'menuTrigger')}}
    ...attributes
  />
  <menu.Items
    {{popper-tooltip this.menuTrigger}}
  />
</Menu>
```

If `Menu` creates a DOM node of it's own, I probably need _some_ way to add classes to both `Menu` _and_ the menu trigger since that might end up being necessary to get the layout where this component is used correct. Needing to think about both of these nodes is unnecessary overhead from the perspective of someone making use of`Menu`. If I _want_ that wrapper node, it's easy to wrap this whole thing in a `div`, but I bet in most cases it would be better to conceptually map my component to a single DOM node, the menu trigger.

alexlafroscia:
I'd love to get a review from @achambers as well before merging, just to make sure everyone's on the same page with the changes here since it's a breaking change!

GavinJoyce:
> I'm thinking about a case where someone builds out a component that wraps Menu. If it were me, I would do something like this

👍 I'm convinced, thanks

achambers (approved):
Love your work @alexlafroscia. I'm in support of this improvement. Thanks.

alexlafroscia:
Thanks @achambers !

PR #59, opened 2021-06-26
Set `aria` relationship between Menu and Menu Items

alexlafroscia:
The core intention here is to set up the `aria-activedescendent` property on the `Menu::Items` component. I initially noticed it was missing while implementing #57, but noticed that the _existing_ tests will look for this attribute, even though it never existed in the first place! (Thankfully, all of those tests still pass now that the attribute actually _is_ supported).

The tests have also been updated so that, in cases where we check for which menu item is active, that check uses this `aria-activedescendent` property. This has a number of benefits:

1. Mirror the way the React/Vue tests are written
2. Less test markup, by asserting against attributes that are part of the implementation (rather than adding additional `data-test` attributes that can signal the same things
3. Ensure that the accessibility implementation is correct by using it to write the tests

### DIFF
--- a/addon/components/menu/items.hbs
+++ b/addon/components/menu/items.hbs
@@ -3,6 +3,7 @@
     id={{@guid}}
     aria-labelledby={{@buttonGuid}}
     tabindex='-1'
+    role='menu'
     ...attributes
     {{on 'keydown' this.onKeydown}}
   >
--- a/addon/components/menu/items.hbs
+++ b/addon/components/menu/items.hbs
@@ -2,6 +2,7 @@
   <div
     id={{@guid}}
     aria-labelledby={{@buttonGuid}}
+    aria-activedescendant={{@activeItem.element.id}}
     tabindex='-1'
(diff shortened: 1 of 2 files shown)


addon/components/menu/items.hbs
```diff
@@ -2,7 +2,9 @@
   <div
     id={{@guid}}
     aria-labelledby={{@buttonGuid}}
+    aria-activedescendant={{this.activeDescendantId}}
```
GavinJoyce, line 5: perhaps we could avoid the indirection and do:

`aria-activedescendant=@activeItem.element.id`

alexlafroscia: 😅 I had that originally and replaced it before actually testing it out, thinking it wouldn't work if `@activeItem` isn't defined. Property access in templates automatically go through `Ember.get` though (IRRC) so that should be safe after all!

GavinJoyce (approved):
🥇

PR #64, opened 2021-07-05
✨ `<Dialog/>` feature/support for nesting dialogs

### LINKED ISSUE #63: ✨ `<Dialog>`Add support for nested dialogs
### **Summary**
As `react` and `vue` packages support nested dialogs, we should introduce the same concept.

[example in react](https://codesandbox.io/s/headlessui-dialog-multiple-modals-bugs-forked-o2ykb?file=/pages/index.js)

The concept assumes that a parent dialog cannot be closed when at least one child dialog is opened. By clicking outside (ESC, on overlay) we only close the dialog which was opened lastly (leaf).

### **Detailed design**

To achieve this goal, we need sort of `counter` / `state` in each dialog, which will keep track of information that at least one children dialog is opened. For communication between parent - child, vue uses concept of inject/provider, [link](https://v3.vuejs.org/guide/component-provide-inject.html).
In practice they don't have to pass explicitly properties to child component in template. 
As in ember we don't have similar mechanism, and we would like to keep headless API as much unified as possible, I propose to use yield emberism. We can yield another value from `<Dialog>` component, e.g. `Nested` and then we can pass `this.element` to its child. Example:

```js
{{yield (hash
   isOpen=@isOpen
   onClose=@onClose
   Overlay=(
    component "dialog/-overlay"
    ...
   )
   ...       
   Nested=(component "dialog" parentDialog=this)
)}}
```
then nested dialog use:

```js
<Dialog
      @isOpen={{state.isOpen}}
      @onClose={{set state.isOpen false}}
      as |dialog|
 >
  Parent dialog
  <dialog.Nested>
     child dialog
  </dialog.Nested>
</Dialog>
```

### **Drawbacks**

API is not 100% match, compare to `react` and `vue` where they used just `<Dialog>` inside another `<Dialog>`

### **Alternatives**

I was thinking about service injected to all dialogs, then when a component is mounted, the dialog is registered in service. Each time when new dialog is registered, the service iterates on opened dialogs and checks relationship parent - child. Open question is: as long as we use portals (dialogs are render as siblings) we cannot rely on DOM nodes, native events for checking parent - child. @achambers you said about some solution for this problem - I cannot remember details - can you remind me ?

@GavinJoyce @alexlafroscia 
Would be wonderful to see your voices here.

far-fetched:
Fixes #63

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 10.x
+          node-version: 14.17.2
 
       - uses: actions/cache@v2
         with:
--- a/tests/integration/components/dialog-test.js
+++ b/tests/integration/components/dialog-test.js
@@ -1,8 +1,13 @@
 import { module, test, todo } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { click, render, setupOnerror } from '@ember/test-helpers';
-import triggerKeyEvent from '@ember/test-helpers/dom/trigger-key-event';
+import {
+  click,
+  render,
+  setupOnerror,
+  triggerKeyEvent,
+} from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
+import userEvent from '@testing-library/user-event';
 
 import { Keys } from 'ember-headlessui/utils/keyboard';
 
@@ -13,7 +18,9 @@ import {
   assertDialogOverlay,
   assertDialogTitle,
   getDialog,
+  getDialogs,
   getDialogOverlay,
+  getDialogOverlays,
   assertActiveElement,
   getByText,
 } from '../../accessibility-assertions';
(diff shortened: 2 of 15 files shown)


tests/integration/components/dialog-test.js
```diff
@@ -1,8 +1,13 @@
...
-import { click, render, setupOnerror } from '@ember/test-helpers';
-import triggerKeyEvent from '@ember/test-helpers/dom/trigger-key-event';
+import {
+  click,
+  render,
+  setupOnerror,
+  triggerKeyEvent,
+} from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
+import userEvent from '@testing-library/user-event';
```
far-fetched, line 10: As `triggerEvent` and `triggerKeyEvent` don't simulate changing focus element - we need some emulator.
In `react` they created own script:
[link](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/test-utils/interactions.ts#L291)

achambers (approved):
Oh, I love this. Feels like the stack service implementation is a good fit.

Love your work. Thanks @far-fetched .

@GavinJoyce @alexlafroscia Fancy giving this a looksee before I merge and ship it?

achambers:
@far-fetched Obviously the broken test run needs fixing. Node version update needed by the looks of it.

tests/integration/components/dialog-test.js
achambers: I've not used `@testing-library`. This is preferable over re-implementing the same code that you linked to in the headless react lib?

far-fetched: Not sure, but regarding that is only dev dependency, I use it

tests/integration/components/dialog-test.js
```diff
@@ -484,19 +491,274 @@ module('Integration | Component | <Dialog>', function (hooks) {
   });
 
   module('Nesting', function () {
-    todo(
-      'it should be possible to open nested Dialog components and close them with `Escape`',
-      async function () {}
-    );
+    test.each(
```
alexlafroscia, line 494: Nice use of `test.each` here!

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 10.x
+          node-version: 14.17.2
```
alexlafroscia, line 19: What made this required? The `@testing-library` packages?

addon/services/stack-provider.js
```diff
@@ -0,0 +1,22 @@
+import Service from '@ember/service';
+import { action } from '@ember/object';
+
+export default class StackProviderService extends Service {
```
alexlafroscia, line 4: I might think about a more specific name for this service, since it's going to be registered into the host app's global service namespace; I could see someone seeing `stack-provider` in there and asking themselves _what the heck is that?_

addon/services/stack-provider.js
```diff
@@ -0,0 +1,22 @@
...
+  stack = [];
+
+  @action
+  hasOpenChild(dialog) {
```
alexlafroscia, line 8: This is checking that the `dialog` argument is the currently-active dialog, yeah?

The name of this method is throwing me off a little -- it sort of suggests that it's checking if the `dialog` argument has a child that is open, rather than that the `dialog` argument is _itself_ open.

tests/integration/components/dialog-test.js
alexlafroscia: This does feel a little odd to me too, though I honestly can't determine if it's actually _is_ weird or if it's just not a pattern that I've seen before. These packages are meant to be framework agnostic, so it should be fine, I think? 

Maybe it makes more sense to borrow that helper function from the React implementation and use that ourselves, and avoid the extra testing packages?

addon/services/stack-provider.js
far-fetched: Hmm I think that this method does exactly what you said: 

> it sort of suggests that it's checking if the `dialog` argument has a child that is open

if a dialog has an active (opened) child - which is the last element added to the stack, it will return false, otherwise true;

.github/workflows/ci.yml
far-fetched: Yep, then I updated it to last LTS version.

addon/services/stack-provider.js
achambers: Agree. Also consider adding something in there about it being the stack provider for dialogs? `Dialog` is pretty generic and likely to cause a clash anyway (not much we can do about that), so how about `DialogStackProvider`?

addon/services/stack-provider.js
achambers: Maybe turning it around to be something more like: `isCurrentlyOpenDialog` which checks that the provided dialog uid _is_ the last one in the stack? This might be less cognitive load.

far-fetched: hmm this is the oposit logic, so it will change:
```js
if (this.dialogStackProvider.isCurrentlyOpenDialog(this)) {
   this.args.onClose();
}
```

But If 3 dialogs are open (nested dialogs) ? The name of this method does not make sense, right ? I don't fully get what is wrong with `hasOpenChild` 🤔

tests/integration/components/dialog-test.js
far-fetched: It should work with any framework. 
About copying and pasting from react, regarding that they use `jest` there are some framework specific code:
[link](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/test-utils/interactions.ts#L160)
so copy and paste will not work in our work.

addon/services/stack-provider.js
alexlafroscia: My confusion with `hasOpenChild` is that is suggests that the `dialog` argument passed to it is not itself the top-most dialog. It reads to me like the method is checking that the `dialog` being passed in itself has some child dialog, which is open, but I don't think that's the intention.

It might be more clear to adopting the naming of typical Stack operations: `push`, `pop` and `peek`. Then if you wanted to check if a given dialog is the top-most one, you can do this instead:

```javascript
if (this.dialogStackProvider.peek() === this) {
  this.args.onClose()
}
```

The benefit here is that there's no longer ambiguity around what the intention of the code is; `peek` returns the top of a Stack, and we only want to operate if we're working with the `Dialog` at the top of the stack. The service then also conceptually makes to being just a stack without really needing any domain-specific behaviors, which further simplifies things.

PR #68, opened 2021-07-07
Implement `Transition` Component

### LINKED ISSUE #65: Implement `Transition` component
Just creating this as a way of letting people know I am working on this one in my free time, to avoid two people starting on it without knowing the other has something in progress!

alexlafroscia:
This is a first attempt as implementing the `Transition` component from Headless UI. The API has been translated as closely as possible to the React/Vue implementations.

# Component API

We have a top-level `Transition` component that takes the same props that the React/Vue implementations do

```hbs
<Transition
  @show={{this.isShown}}
  @enter="transition transition-duration-100"
  @enterFrom="opacity-0"
  @enterTo="opacity-100"
>
  Show me!
</Transition>
```

Based on the `@show` argument, the underlying element (and any children) will be mounted/unmounted. The "secret sauce" of the component is that when the component becomes _not_ shown, we don't immediately unmount the component; we instead apply the `leave` classes and wait for the transition to complete, _then_ unmount.

We also have a "child" component that can be used to coordinate multiple related transitions

```hbs
<Transition @show={{this.isShown}} as |t|>
  <t.Child 
    @enter="transition transition-duration-100"
    @enterFrom="opacity-0"
    @enterTo="opacity-100"
  >
    Show me!
  </t.Child>

  <t.Child 
    @enter="transition transition-duration-200"
    @enterFrom="opacity-0"
    @enterTo="opacity-100"
  >
    Show me!
  </t.Child>
</Transition>
```

The "secret sauce" with coordinating transition/child components is that unmounting will wait for _all_ components in the group -- the parent and all children -- have completed their transition duration.

# Implementation Trade-Offs

Two additional `dependencies` have been added to the addon to implement `Transition`

* `tracked-maps-and-sets`
  * This allows us to use a `Set` to store the list of applied classes, which a nice data structure to use for this. By using the "tracked" version of `Set`, the UI will automatically re-create the transition class string based on changes to the data.
  * I don't think this dependency can really be removed easily
* ~~`ember-could-get-used-to-this`~~
  * Ember doesn't have a concept of a `Resource` as part of the "official" API. A `Resource` allows us to monitor some tracked properties and produce a value that has it's own lifecycle that reacts to those property changes. This was just the thing we needed to react to the `@show` argument changing and, when becoming `false`, delay the unmounting of the component until after the transition completes
  * The `did-update` modifier _sort of_ helps with this, but because the DOM node for `Transition` isn't always rendered, it's not something we can necessarily rely on to react to the changing value. I'm happy to attempt a refactor to remove the use of a `Resource`, but the implementation might end up being more complicated

I was able to refactor out the usage of `ember-could-get-used-to-this` by leveraging the `invokeHelper` API. The implementation is slightly less "clean", as the use of `Helper` as a base class for these stateful value representations is a little more awkward than the `Resource` base class from `ember-could-get-used-to-this`, but I think it's worthwhile in order to avoid bringing in that whole dependency if it's possible to get it working without it.

We could always implement an in-house `HelperManager` for these two `Helper` implementations that would give us an API closer to that of `Resource` _without_ needing to use `ember-could-get-used-to-this`, but that could be a refactor later as it's not really necessary and does result in more code.

I also added a few `devDependencies` to make testing the events that `Transition` emits easier

- `testdouble` allows us to create stub functions and verify that they were called
- `testdouble-qunit` allows us to integrate QUnit and `testdouble` together by adding an assertion to QUnit for verifying that a `testdouble` stub was called

# To-Do

- [ ] Create documentation pages for the `Transition` component
- [ ] Determine how we feel about the additional dependencies
- [x] `waitForTransition` tests might be flakey; it passes consistently when run locally, but has failed at least once in CI

Closes #65

### DIFF
--- a/addon/components/transition.js
+++ b/addon/components/transition.js
@@ -1,0 +1,88 @@
+import Component from '@glimmer/component';
+import { getValue } from '@glimmer/tracking/primitives/cache';
+import { invokeHelper } from '@ember/helper';
+import { action } from '@ember/object';
+import { schedule } from '@ember/runloop';
+import { assert } from '@ember/debug';
+import {
+  AppliedClassNamesManager,
+  TransitionVisibilityManager,
+} from '../helpers/transition';
+
+export default class TransitionComponent extends Component {
+  constructor() {
+    super(...arguments);
+
+    let { show } = this.args;
+
+    assert(ARG_SHOW_MISSING_ERROR_MESSAGE, typeof show === 'boolean');
+
+    schedule('afterRender', () => {
+      this.initialRender = false;
+    });
+  }
+
+  get tagName() {
+    return this.args.tagName ?? 'div';
+  }
+
+  get unmount() {
+    return this.args.unmount ?? true;
+  }
+
+  /* === Component Visibility Management === */
+
+  allTransitionDomNodes = new Set();
+
+  transitionVisibilityManager = invokeHelper(
+    this,
+    TransitionVisibilityManager,
+    () => ({
+      positional: [this.allTransitionDomNodes, this.args.show],
+    })
+  );
+
+  get componentIsVisible() {
+    return getValue(this.transitionVisibilityManager);
+  }
+
+  get componentIsMounted() {
+    return this.unmount ? this.componentIsVisible : true;
+  }
+
+  get componentIsHidden() {
+    return this.unmount ? false : !this.componentIsVisible;
+  }
+
+  @action
+  trackDomNode(element) {
+    this.allTransitionDomNodes.add(element);
+  }
+
+  @action
+  untrackDomNode(element) {
+    this.allTransitionDomNodes.delete(element);
+  }
+
+  /* === Transition Class Name Management === */
+
+  initialRender = true;
+
+  /** @type {HTMLElement} */
+  ownDomNode;
+
+  appliedClassNamesManager = invokeHelper(
+    this,
+    AppliedClassNamesManager,
+    () => ({
+      positional: [this, this.args.show],
+    })
+  );
+
+  get transitionClassNames() {
+    return getValue(this.appliedClassNamesManager);
+  }
+}
+
+export const ARG_SHOW_MISSING_ERROR_MESSAGE =
+  'You have to provide a `show` prop to the `Transition` component';
--- a/addon/components/transition/child.hbs
+++ b/addon/components/transition/child.hbs
@@ -1,0 +1,20 @@
+<Transition
+  @show={{@show}}
+  @unmount={{@unmount}}
+  @tagName={{@tagName}}
+  @enter={{@enter}}
+  @enterFrom={{@enterFrom}}
+  @enterTo={{@enterTo}}
+  @leave={{@leave}}
+  @leaveFrom={{@leaveFrom}}
+  @leaveTo={{@leaveTo}}
+  @beforeEnter={{@beforeEnter}}
+  @afterEnter={{@afterEnter}}
+  @beforeLeave={{@beforeLeave}}
+  @afterLeave={{@afterLeave}}
+  {{did-insert @parent.trackDomNode}}
+  {{will-destroy @parent.untrackDomNode}}
+  ...attributes
+>
+  {{yield}}
+</Transition>
--- a/addon/components/transition.hbs
+++ b/addon/components/transition.hbs
@@ -1,0 +1,21 @@
+{{#if this.componentIsMounted}}
+  {{#let (element this.tagName) as |Tag|}}
+    <Tag
+      class={{this.transitionClassNames}}
+      hidden={{this.componentIsHidden}}
+      style={{if this.componentIsHidden 'display: none'}}
+      {{did-insert (set this 'ownDomNode')}}
+      {{did-insert this.trackDomNode}}
+      {{will-destroy this.untrackDomNode}}
+      ...attributes
+    >
+      {{yield
+        (hash
+          Child=(component
+            'transition/child' parent=this show=@show unmount=@unmount
+          )
+        )
+      }}
+    </Tag>
+  {{/let}}
+{{/if}}
--- a/addon/components/transition/child.js
+++ b/addon/components/transition/child.js
@@ -1,0 +1,15 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+export default class TransitionChildComponent extends Component {
+  constructor() {
+    super(...arguments);
+
+    let { parent } = this.args;
+
+    assert(PARENT_MISSING_ERROR_MESSAGE, typeof parent !== 'undefined');
+  }
+}
+
+export const PARENT_MISSING_ERROR_MESSAGE =
+  '<Transition::Child /> is missing a parent <Transition /> component.';
--- a/addon/helpers/transition/applied-class-names.js
+++ b/addon/helpers/transition/applied-class-names.js
@@ -1,0 +1,141 @@
+import Helper from '@ember/component/helper';
+import { dropTask, timeout } from 'ember-concurrency';
+import {
+  ClassNameSet,
+  splitClassNames,
+  waitForTransition,
+} from '../../utils/transition';
+
+export class AppliedClassNamesManager extends Helper {
+  appliedClassNames = new ClassNameSet();
+
+  /**
+   * Tracks the last `this.show` value, so we can determine if we need to initiate a transition as part of `compute`
+   * being executed
+   *
+   * This is because `compute` is sometimes called because an argument changed (and we want to kick off a new
+   * transition) and sometimes it's called to update the produced value during an existing transition (and we *do
+   * not* want to kick off a new transition)
+   */
+  previousShowValue = false;
+
+  /** @type {import('../../components/transition').default} */
+  get transition() {
+    return this.args[0];
+  }
+
+  /**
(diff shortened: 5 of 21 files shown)


alexlafroscia:
So, when I went to actually _use_ the `Transition` component in the docs site... I realized it didn't really work 😅 

After a bit of refactoring, it became clear that the actual class string that should be applied to the element is _also_ a great candidate for a `Resource`. The previous attempt, which used both `did-insert` and `did-update`, didn't actually apply the classes correctly when entering the component.

I think that using a `Resource` for both of these things makes the implementation more clear as well, though it _does_ further couple things to `ember-could-get-used-to-this`. Worth the trade-off? Not sure yet...

addon/components/transition.js
```diff
@@ -0,0 +1,72 @@
...
+    super(...arguments);
+
+    let { show } = this.args;
+
+    assert(ARG_SHOW_MISSING_ERROR_MESSAGE, typeof show === 'boolean');
+
+    // Normally this value is computed lazily, but we *do not want this*
+    // If we allow the value to be lazy, the class names are first computed only once the component is rendered
+    // Simply accessing the property will cause the resource to begin computing
+    this.transitionClassNames;
```
alexlafroscia, line 22: I want to play around with the implementation some more and see if this can be avoided.

The tricky bit here is that we want the Resource that backs the class name string to start working right away, as soon as the component is rendered. This is so that when we change `@show` from `false` to `true`, we can respond to that and perform the transition as expected.

When we allow this value to be lazy, we lose the ability to respond to the transition; from the perspective of the Resource, it appears that we're performing our initial render with a `@show` value that is `true`, when in reality the initial render took place earlier with a `@show` value of `false`.

The Resource is initialized only once `@show` becomes `true` for the first time because, without this code that is eagerly forcing it to be instantiated, we only access the class string the first time we actually render the component contents for the first time.

An alternative that I want to explore is having the `Transition` component monitor whether we're in our first render or not. This context can be passed along to the Resource.

addon/components/transition.js
```diff
@@ -0,0 +1,72 @@
...
+    return this.args.tagName ?? 'div';
+  }
+
+  get unmount() {
+    return this.args.unmount ?? true;
+  }
+
+  /* === Component Visibility Management === */
+
+  @tracked domNodes = new Set();
```
alexlafroscia, line 35: This, and the `domNode` on line 63, need better names

This property contains all of the DOM nodes related to the entire `Transition` tree (including children)

The other property is specifically the DOM node for _this_ `Transition` component

addon/components/transition/child.hbs
```diff
@@ -0,0 +1,20 @@
...
+  @enter={{@enter}}
+  @enterFrom={{@enterFrom}}
+  @enterTo={{@enterTo}}
+  @leave={{@leave}}
+  @leaveFrom={{@leaveFrom}}
+  @leaveTo={{@leaveTo}}
+  @beforeEnter={{@beforeEnter}}
+  @afterEnter={{@afterEnter}}
+  @beforeLeave={{@beforeLeave}}
+  @afterLeave={{@afterLeave}}
```
alexlafroscia, line 14: I wish there was a better way to pass along all of the arguments, but since there isn't, we need to manually forward all of the arguments to the inner component!

alexlafroscia:
Thoughts on landing this without “documentation”?

Considering the way that we’re planning to re-think how the docs site looks/works before 1.0, it might not be SO important to land real docs as part of this PR. There’s an example of using `Transition` with `Menu` (an example that was previously marked as a “todo”) so there’s the test suite and _that_, at least.

achambers:
Woah, this is seriously cool @alexlafroscia. Fantastic work man. I won't pretend to know exactly what's going on. You've clearly put a lot of thought and effort in to this. Very cool.

In terms of the documentation, makes sense. What are the current thoughts on re-thinking the docs before 1.0?

In terms of dependencies, so you removed the `ember-could-get-used-to-this` I see. Something about the experimental nature of that addon feels like that's a good idea.

As for `tracked-maps-and-sets` 👍 . That's a great addon. I support it's inclusion 😄 

One thing I'm curious about.....that ember-css-transitions addon clones the element for the purpose of leave transitions. We don't seem to do that. Can you help me understand what we do instead?

Also, one thing I noticed when I was playing with ember-css-transitions was that if a DOM element had leave transitions, and that DOM element was within another element and that parent element is torn down by Ember, the leave transitions never had an opportunity to run. I think because Ember removed the parent element from the DOM which included the child element that had the transitions. Does this implementation cater for that or is that something that the consumer needs to be aware of?

alexlafroscia:
> One thing I'm curious about.....that ember-css-transitions addon clones the element for the purpose of leave transitions. We don't seem to do that. Can you help me understand what we do instead?

Here's my understanding: at the end of the day, to do a CSS "out" transition the way that the `Transition` component and `ember-css-transitions` does you need to disconnect the declarative "my element is rendered" state from the "the element is actually in the DOM" state. This is so that the transition from "rendered" to "not rendered" can first have the CSS classes applied.

One approach is cloning the element and manually re-attaching the clone after Ember has stopped rendering the component, so that Ember's state can say "this element is not rendered" but the DOM node _actually_ hangs out some amount of time to perform the transition out.

We're taking a different approach here that simply disconnects the "you should be visible" argument from the "you actually _are_ visible" internally property. We hook into the transition of the "you should be visible" argument to set a timeout (using Ember Concurrency) that, once expired, changes the "you actually _are_ visible" state to match. At the point that that timeout starts, we apply our "transition out" classes so that whatever should happen animation wise can happen.

> Does this implementation cater for that or is that something that the consumer needs to be aware of?

The latter. The transition animations are applied based off the `show` argument changing, _not_ the `Transition` component being inserted or removed. I don't think that the "real" HeadlessUI does anything to handle that case and anything we might _try_ to do here would be a major hack. I think it's better to keep this simple and have people work with the given API here. Trying to animate out the `Transition` component actually un-mounting opens pandora's box: for example, do we need to delay route transitions for animation purposes? That's a road I'd rather not go down!

GavinJoyce:
This looks amazing, nice one @alexlafroscia!

I don't currently have time to review it properly, glad that @achambers has been able to review it somewhat.

> Thoughts on landing this without “documentation”?

👍 sounds good to me

achambers:
> > One thing I'm curious about.....that ember-css-transitions addon clones the element for the purpose of leave transitions. We don't seem to do that. Can you help me understand what we do instead?
> 
> Here's my understanding: at the end of the day, to do a CSS "out" transition the way that the `Transition` component and `ember-css-transitions` does you need to disconnect the declarative "my element is rendered" state from the "the element is actually in the DOM" state. This is so that the transition from "rendered" to "not rendered" can first have the CSS classes applied.
> 
> One approach is cloning the element and manually re-attaching the clone after Ember has stopped rendering the component, so that Ember's state can say "this element is not rendered" but the DOM node _actually_ hangs out some amount of time to perform the transition out.
> 
> We're taking a different approach here that simply disconnects the "you should be visible" argument from the "you actually _are_ visible" internally property. We hook into the transition of the "you should be visible" argument to set a timeout (using Ember Concurrency) that, once expired, changes the "you actually _are_ visible" state to match. At the point that that timeout starts, we apply our "transition out" classes so that whatever should happen animation wise can happen.
> 
> > Does this implementation cater for that or is that something that the consumer needs to be aware of?
> 
> The latter. The transition animations are applied based off the `show` argument changing, _not_ the `Transition` component being inserted or removed. I don't think that the "real" HeadlessUI does anything to handle that case and anything we might _try_ to do here would be a major hack. I think it's better to keep this simple and have people work with the given API here. Trying to animate out the `Transition` component actually un-mounting opens pandora's box: for example, do we need to delay route transitions for animation purposes? That's a road I'd rather not go down!

Wowsers. Thanks for the detailed description @alexlafroscia. Super useful. Outstanding. I'm going to have another pass at the code and going to pull it down to have a play with a few transitions today. But it's looking really great mate. Super happy how this is playing out. Stellar work 🙌

alexlafroscia:
Thanks @achambers! Eager to land this one 🚀

achambers:
> Thanks @achambers and @GavinJoyce! Eager to land this one 🚀

Sorry mate, got delayed today. Will take a look tomorrow.

alexlafroscia:
No rush at all!

alexlafroscia:
Just wanted to give this a bump; there is some refactoring in my app at work that I'd love to do using this component, so I finally have a vested interest in landing it 😅 

If something like a code-tour to review it together would be helpful, I'd be happy to do something like that!

achambers:
> Just wanted to give this a bump; there is some refactoring in my app at work that I'd love to do using this component, so I finally have a vested interest in landing it 😅
> 
> If something like a code-tour to review it together would be helpful, I'd be happy to do something like that!

Hey @alexlafroscia. I've been away on holiday, hence my silence. I'd love a code-tour on this. Today is tricky but any other day this week works for me. Let me know when suits and we can jump on a call.

alexlafroscia:
Any time this week is going to be tough for me, unfortunately... I'm in the last month before my wedding and between finally preparation, family in town, etc. my calendar is crazy these days!

I just sent you a friend request on Discord @achambers -- maybe we can sort out a time that would work for us to review this there!

alexlafroscia:
Talked to @GavinJoyce offline (on Discord); he's comfortable merging this and circling back on an in-depth review later, given limited time to do a deep code review

PR #70, opened 2021-08-13
Break out CI testing jobs

alexlafroscia:
I noticed that the tests seem to be failing on `master`, but it wasn't immediately clear _why_. This is in part because the test command does 6 things all at once:

- lint
- normal `ember test`
- runs Ember Try (4 different scenarios)

While this doesn't _fix_ the failures right now, it _does_ make it a lot more clear what exactly is going wrong by identifying each of those 6 jobs independently. As a bonus, we only bother running Ember Try if the tests pass under normal circumstances!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,4 +53,32 @@ jobs:
         run: yarn install --frozen-lockfile --non-interactive
 
       - name: test
-        run: yarn test
+        run: yarn ember test
+
+  ember-try:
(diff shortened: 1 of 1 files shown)


GavinJoyce (approved):
👍 nice one @alexlafroscia

alexlafroscia:
Even though the tests are failing, I feel like this isn't _worse_ than the current state and makes it more clear to a potential consumer what specifically isn't working right now.

PR #71, opened 2021-08-13
Fix Failing CI Tests

alexlafroscia:
It seems like CI is failing due to a deprecation warning (error?) stemming from tests that mutate the result of the `hash` helper.

This should fix the tests!


GavinJoyce (approved):

alexlafroscia:
Clearly it wasn't that that caused the failures 😅 I'll keep iterating in this PR until we can get the tests "green" again!

alexlafroscia:
It seems like maybe the `(hash)` mutation _was_ the issue, I just didn't get all of the instances the first time 😅 I ran the `ember-beta` case locally now and it seems like it's working!

PR #72, opened 2021-08-15
Enable Embroider Tests

alexlafroscia:


alexlafroscia:
The current failures aren't happening locally... I wonder what's happening here!

CleanShot 2021-08-14 at 21 43 16@2x

alexlafroscia:
I keep trying to get this working but can't understand what's going wrong...

I did a little more digging into what `ember-try` _specifically_ does under-the-hood while running a scenario. It mentions installing without the lockfile, to automatically grab the latest dependencies, so I tried that locally and _still_ cannot reproduce the specific exception that is triggered in CI here that ends up hanging the job.

NullVoxPopuli:
Anything preventing this from landing? I see the build is green

PR #76, opened 2021-08-31
✨ Implement `<Listbox />` component

### LINKED ISSUE #69: Implement `<Listbox>` component
I have already implemented it in one of my pet project. There is a lack of accessibility features there, but at least it would be a good starting point. I just let you know to not repeat work. I can push it forward in spare time.

dmcnamara-eng:
###  ✨ What Changed and Why

This PR is a functionally complete implementation of a `<Listbox />` component based on https://headlessui.dev/react/listbox

Things that this PR implements:

* Show/hide options on mouse click
* Support for disabled component and individual options
* Support for a optional label element
* Full mouse support for focus and active elements
* Full keyboard support (`Enter`/`Space`/`ArrowUp`/`ArrowDown`/`ArrowLeft`/`ArrowRight`/`PageUp`/`PageDown`/`Home`/`End`/`Escape`)
* Support for searching the list using the keyboard
* Click outside the listbox to close
* Full accessibility implementation (`aria-expanded`/`aria-labelledby`/`aria-controls`/`aria-haspopup`/`aria-active-descendent`) and focus management
* Sample implementation in dummy app

Things that this PR does not implement are:

* static prop
* Composition with transitions

I ported the [React component tests for listbox](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/listbox/listbox.test.tsx#L28) as Ember integration tests. 90 integration tests were ported. Some tests were not ported because they are not applicable to Ember components. A test for transitions was not ported because transitions has not been implemented here yet.

### 😈 Before

Didn't exist.

### 😇 After

after

Closes #69


addon/components/listbox.hbs
```diff
@@ -0,0 +1,46 @@
+<div
+  {{did-insert this.registerDocumentClickListener}}
+  {{will-destroy this.unregisterDocumentClickListener}}>
+
+  {{yield
+    (hash
+      open=this.isOpen
+      disabled=(if @disabled true false)
```
alexlafroscia, line 8: It seems like we define `disabled` here differently than we do when it's passed into the `Label` component on line 39. It would be good to be consistent about what `disabled` represents. Maybe we should have a property on the `Listbox` component instance that handles normalizing the argument into a Boolean, which we can reference with confidence everywhere we need access to that?

addon/components/listbox.js
```diff
@@ -0,0 +1,316 @@
...
+      optionId = optionComponent.guid + '-option';
+      optionValue = optionComponent.args.value;
+    } else if (this.activeOptionIndex !== undefined) {
+      optionId = this.optionElements[this.activeOptionIndex].id;
+      optionValue = this.optionValues[this.activeOptionIndex];
+    } else {
+      return;
+    }
+
+    this.optionElements.find((o, i) => {
```
alexlafroscia, line 175: I think there are a few opportunities for refactoring with this functionality here

- We're using `.find`, which returns a specific member of `optionElements`, but don't actually ever _use_ the value that it's returning
- We're keeping track of whether the element has been located with a separate value, where that shouldn't be necessary; "finding" an element is already what `.find` does
- We're performing side-effects within `.find`, which isn't necessarily _wrong_ but is an unusual way of using the function

I would use `.find` to locate an element and then move all of the usage of the located element outside of the `.find` callback instead. We can simplify the code and remove the `found` state variable that way!

addon/components/listbox/-button.hbs
```diff
@@ -0,0 +1,18 @@
+{{#let (element (or @as 'button')) as |Tag|}}
```
alexlafroscia, line 1: The other components so far have used `@tagName` to set an alternate element that can be used as root element; I agree that `@as` is a better name, but for now it makes more sense to be consistent with the usage of `@tagName` and migrate them _all_ to use `@as` down the road

addon/components/listbox/-button.hbs
```diff
@@ -0,0 +1,18 @@
...
+  <Tag
+    role={{if @role @role 'button'}}
```
alexlafroscia, line 3: It doesn't seem like the upstream `Listbox.Button` component has an `@role` argument; is it necessary for us to support it?

addon/components/listbox/-button.hbs
```diff
@@ -0,0 +1,18 @@
...
+    id='{{@guid}}-button'
+    aria-haspopup='listbox'
+    aria-controls={{@guid}}
+    aria-labelledby='{{@guid}}-label {{@guid}}-button'
+    {{on 'click' @handleButtonClick}}
+    {{on 'keydown' @handleKeyDown}}
+    {{on 'keypress' @handleKeyPress}}
+    {{aria "expanded" value=@isOpen removeWhen=@disabled}}
```
alexlafroscia, line 11: I don't think this `{{aria}}` helper is necessary; if the attribute is set to a false-y value, it will be removed from the DOM node. We should be able to create an appropriate value that we can set `aria-expanded` to to avoid needing this custom behavior

addon/components/listbox/-option.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { guidFor } from '@ember/object/internals';
+
+export default class ListboxOptionComponent extends Component {
+  @tracked guid = `${guidFor(this)}-headlessui-listbox`;
```
alexlafroscia, line 6: `@tracked` can be removed here, since the value will not change after being set in the constructor for this class

addon/components/listbox/-options.hbs
```diff
@@ -0,0 +1,27 @@
+{{#if @isOpen}}
+  <ul
```
alexlafroscia, line 2: It would be good to support changing the tag name used here, as we can with the other components by setting an `@tagName` argument

addon/components/listbox/-options.hbs
```diff
@@ -0,0 +1,27 @@
...
+    tabindex="-1"
+    role='listbox'
+    aria-labelledby={{if @hasLabelElement (concat @guid '-label') (concat @guid '-button')}}
+    aria-activedescendant={{@activeOptionGuid}}
+    {{did-insert @registerOptionsElement}}
+    {{will-destroy @unregisterOptionsElement}}
+    {{on 'keypress' @handleKeyPress}}
+    {{on 'keydown' @handleKeyDown}}
+    ...attributes
+    class="pb-1"
```
alexlafroscia, line 13: These components shouldn't contain any `class` values; visual styling should be left up to the user!

tests/integration/components/listbox-test.js
```diff
@@ -0,0 +1,3354 @@
...
+      // Verify it is closed
+      assertListboxButton({ state: ListboxState.InvisibleUnmounted });
+      assertListbox({ state: ListboxState.InvisibleUnmounted });
+
+      // Verify the button is focused again
+      await assertActiveElement(getListboxButton());
+    });
+
+    test('should be possible to close the listbox with Enter and choose the active listbox option', async function () {
+      let handleChange = sinon.spy();
```
alexlafroscia, line 783: We already have `testdouble` installed for creating spys and asserting their use; I'd rather we stick to one dependency of that variety than using two different ones in different test modules

alexlafroscia (commented):
I'm really excited to see this PR!

I took a first pass through and left some thoughts; there is a _lot_ to review here so it might take me a few passes.

 I'm generally going to assume the test implementation is fine and that the tests passing are enough of an indication that they're good-to-go, and will instead focus on the actual add-on source code for review purposes (at least, for now!)

dmcnamara-eng:
> I'm really excited to see this PR!
> 
> I took a first pass through and left some thoughts; there is a _lot_ to review here so it might take me a few passes.
> 
> I'm generally going to assume the test implementation is fine and that the tests passing are enough of an indication that they're good-to-go, and will instead focus on the actual add-on source code for review purposes (at least, for now!)

Sounds good! Thanks for the quick feedback! I've replied to individual comments above.

addon/components/listbox.hbs
```diff
@@ -0,0 +1,46 @@
+<div
+  {{did-insert this.registerDocumentClickListener}}
+  {{will-destroy this.unregisterDocumentClickListener}}>
+
+  {{yield
+    (hash
```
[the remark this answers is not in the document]
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/f80e94f920e5dacf03a5d40a464bf19ebb91f219 exposes the `openListbox` and `closeListbox` action to yielded components. 

The `isOpen` property is now a `getter`. This allows an app to toggle `isOpen` and have the listbox component internally keep its state consistent.

addon/components/listbox.hbs
```diff
@@ -0,0 +1,46 @@
...
+      open=this.isOpen
```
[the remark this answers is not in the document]
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/f80e94f920e5dacf03a5d40a464bf19ebb91f219 renames this from `open` -> `isOpen`

addon/components/listbox.hbs
```diff
@@ -0,0 +1,46 @@
+<div
+  {{did-insert this.registerDocumentClickListener}}
```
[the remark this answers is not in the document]
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/8fb0a7f2720a0c0c7d6140591d9bb5b70449f051 uses `ember-focus-trap` to manage focus state. This removes a lot of focus management from within the component.

https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/8fb0a7f2720a0c0c7d6140591d9bb5b70449f051 also employs `click-outside` meaning we don't need to explicitly manage document clicks.

This also removes the extraneous DOM wrapper node.

addon/components/listbox.hbs
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/4eca07ac5429cac6c5a7fb83c6e8de5b1bf5191e implements an internal `isDisabled` property which ensure disabled attribute consistency across the components.

addon/components/listbox.js
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/9779e56ec174ea9f18cf492f122235bf0da05df2 refactors this method to remove the `find` call altogether.

In order to improve performance in large lists, the Listbox component now maintains an index on each option component and element. This avoids use of `find` altogether, as the option can now be referenced by its index in the list.

addon/components/listbox/-options.hbs
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/2489d1156b5cc1016a184acf1a1e92202c0f9a37 allows custom tag names for this component.

I'm using the `@as` property to specify a tag name, as this is consistent with the React API and is also a pattern that `<Dialog>`  follows.

tests/integration/components/listbox-test.js
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/dac162c4e538f6f6bac6648751dcfbbd5949d242 has removed `ember-sinon` as a dependency.

I couldn't find where we use `testdouble` so I just implemented the assertions natively.

addon/components/listbox/-button.hbs
alexlafroscia: It seems that those three states should be, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaExpanded):

1. The string `"true"`
2. The string `"false"`
3. The string `"undefined"`

Maybe we don't need to remove the attribute completely at all?

If we _do_ want to remove attribute entirely in the third case, we can set it to the value `undefined` (rather than `"undefined"`) and the attribute will be removed 

> Don't you think we should follow the same pattern ? I assumed that guys from react attach importance to accessibility.

Absolutely! We just don't need to overcomplicate things to achieve this 😬

addon/components/listbox.hbs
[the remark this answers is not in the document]
alexlafroscia: > exposes the openListbox and closeListbox action to yielded components.

How are they exposed to the user of the `Listbox` component in their template? I can't find where they are `yield`-ed out of the component again

> The isOpen property is now a getter. This allows an app to toggle isOpen and have the listbox component internally keep its state consistent.

Could you point me toward this behavior in the implementation? I can't find a reference to an `open` or `isOpen` argument that the component defers to if provided

addon/components/listbox/-option.js
alexlafroscia: It seems like we only ever reference this `guid` property by appending `-option` to the end of it; would it make more sense to define the `guid` property with that `-option` suffix and avoid needing to remember that each location needs to have the suffix appended manually?

addon/components/listbox.js
```diff
@@ -0,0 +1,306 @@
...
+  @action
+  handleButtonClick(e) {
+    if (e.button !== 0) return;
+    this.activateBehaviour = ACTIVATE_NONE;
+    this.isOpen = !this.isOpen;
+  }
+
+  @action
+  handleClickOutside(e) {
+    if (e.button !== 0) return;
```
alexlafroscia, line 66: What does this check do?

tests/integration/components/listbox-test.js
alexlafroscia: Whoops, that's my bad! I added it in the `Transition` PR thats' been open a while and forgot this hasn't actually landed yet

https://github.com/GavinJoyce/ember-headlessui/pull/68/commits/0a5fefd5b2726d16106bbe45c7e1175453980f41

We could use `ember-sinon` instead of you have a hard preference there, or refactor the tests down the line to use `testdouble` once all of these PRs are merged

addon/components/listbox/-options.hbs
alexlafroscia: Huh, I didn't realize we were _already_ inconsistent about whether to use `@tagName` or `@as`

I like `@as` better anyway, so sticking with that makes sense to me! I'll update the other components to follow that pattern

addon/components/listbox/-options.hbs
alexlafroscia: That's alright; I think the test assertion using `.exists()`, as you refactored it to, is fine for our purposes!

alexlafroscia:
I approved the GitHub Actions run so we'll get lint/test results for the PR now. GitHub disables those for first-time contributors (people were abusing the system to mine cryptocurrency 🤦‍♂️) do you need a contributor to give the OK for CI to run on an individual's first PR!

addon/components/listbox.hbs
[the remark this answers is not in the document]
dmcnamara-eng: Ah sorry, I think I originally misunderstood 😄 

https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/45a2f574d8f1187255c4c0f7b88df447352060b3 has the change to `yield` out `openListbox` and `closeListbox` and also set initial state from a `isOpen` property.

addon/components/listbox/-button.hbs
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/11c9843bce96d08007f499c4e3a66574115e1d6f removes the `aria` modifier.

addon/components/listbox/-option.js
dmcnamara-eng: https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/2c8811db9b9f3a0ea02136f36418607a18748d85 cleans this up, so that the `guid` always has the correct suffix.

dmcnamara-eng:
> Rather than using the `click-outside` modifier on the button, I think we should be able to use the `onDeactivate` option to close the menu and avoid needing to perform any focus-management-on-click-outside behavior ourselves. 

https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/660f5cf28fdee9d2c16ba23f2ba3a2a4205ed3ff removes the `click-outside` modifier to use a pure `focus-trap` implementation.

I had to `skip` [one test](https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/660f5cf28fdee9d2c16ba23f2ba3a2a4205ed3ff#diff-ef3d53d281ee95197b5d4ecbaccd9d1cafb901dcda37b11e630e5fd749d94900R2962) for this to pass. The value of that test seems dubious at this point.

addon/components/listbox.js
dmcnamara-eng: Mostly to make a right-click test pass. It doesn't apply in real world 😆 

It has been removed anyway as part of https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/660f5cf28fdee9d2c16ba23f2ba3a2a4205ed3ff

addon/components/listbox.js
```diff
@@ -0,0 +1,316 @@
...
+        this.open();
+      }
+    } else if (event.key.length === 1) {
+      this.addSearchCharacter(event.key);
+    }
+  }
+  @action
+  handleLabelClick(e) {
+    e.preventDefault();
+    if (e.ctrlKey) return;
```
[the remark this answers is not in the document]
dmcnamara-eng: This has been removed as part of https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/660f5cf28fdee9d2c16ba23f2ba3a2a4205ed3ff

addon/components/listbox.js
```diff
@@ -0,0 +1,298 @@
...
+
+  get selectedOptionGuid() {
+    return this.optionElements[this.selectedOptionIndex]?.id;
+  }
+
+  get isOpen() {
+    return this._isOpen;
+  }
+
+  set isOpen(isOpen) {
```
alexlafroscia, line 40: I really like using a "setter" like this; keeps the code for manipulating the value clean, but allows us to perform these side-effects easily as well!

addon/components/listbox/-option.hbs
```diff
@@ -0,0 +1,26 @@
...
+    tabindex='-1'
+    {{did-insert (fn @registerOptionElement this)}}
+    {{on 'focus' (fn @setActiveOption this)}}
+    {{on 'mouseover' (fn @setActiveOption this)}}
+    {{on 'mouseout' @unsetActiveOption}}
+    {{on 'click' this.handleClick}}
+    aria-selected={{if
+      this.isSelectedOption
+      (if this.isSelectedOption 'true' 'false')
+    }}
```
alexlafroscia, line 14: I'm curious about this conditional; it seems like the only values it will produce are `"true"` or `undefined`. How does `"false"` actually end up returned from this expression?

If `isSelectedOption` is truthy, then both conditionals evaluate to `true` and `"true"` is the result

If `isSelectedOption` is false, then the first conditional will return `undefined` since there isn't an "else" condition in there.

I _think_ the conditional on like 13 here will _always_ return `true` since it'll only even be evaluated if the `this.isSelectedOption` is truthy

alexlafroscia (commented):
This is looking really great! Once we get clarity on the nested conditional I commented on, I am feeling good about merging this PR!

tests/accessibility-assertions.js
```diff
@@ -37,6 +61,12 @@ const DialogState = {
   InvisibleUnmounted: 'InvisibleUnmounted',
 };
 
+const ListboxState = {
+  Visible: 'Visible',
+  InvisibleHidden: 'InvisibleHidden',
```
far-fetched, line 66: I don't see use of this state in tests, I assume that we are just lack of some tests right ? Some assertions don't implement it. Do we have a plan to extend the test suite ?

far-fetched: Maybe would be nice to follow the same approach what @achambers did in dialog ? What I mean is copy test e.g. from react repo and change to `todo` [example](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/dialog-test.js#L205) 
I am willing to help with them after that and fill gaps.

addon/components/listbox/-option.hbs
dmcnamara-eng: You're right, that makes no sense 🤦 

https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/bdce10eeeb8dd5c1ee2ef9fe0c6d2a1cc9ffb750 has a simplified conditional which evaluates to `aria-selected="true"` if the option is selected, or omits the attribute if not selected.

tests/accessibility-assertions.js
dmcnamara-eng: Ah yes this refers to the alternative "hidden" or "unmount" render strategy, in which React will insert the listbox options into the DOM, but display as hidden.

I think it's safe to push this feature out for now.

https://github.com/GavinJoyce/ember-headlessui/pull/76/commits/b452c35fc3e8bb69f110df608dab582f6e148a6d adds `todo`s for these tests.

alexlafroscia (commented):
This is looking for to me! I'm happy to merge it at this point if you're ready @dmcnamara-eng !

alexlafroscia (approved):

dmcnamara-eng:
> This is looking for to me! I'm happy to merge it at this point if you're ready @dmcnamara-eng !

🥳 I'm happy. Please merge!

alexlafroscia:
`0.9.0` released with this PR in it! Awesome work here @dmcnamara-eng 🎉 💯

GavinJoyce:
❤️

miguelcobain:
@dmcnamara-eng how difficult would it be to implement the `multiple` feature on headlessui? https://headlessui.com/react/listbox#selecting-multiple-values

knownasilya:
This renders in place instead of absolutely positioned in the body?

PR #79, opened 2021-09-07
Fix build error for consuming apps, caused by (set) helper

roomman:
My consuming app was throwing the following error, using this addon with the `<Transition>` component:
```
Build Error (PackagerRunner) in node_modules/ember-headlessui/components/transition.hbs

Module Error (from .../node_modules/thread-loader/dist/cjs.js):
Missing helper: set in node_modules/ember-headlessui/components/transition.hbs
```
This PR fixes the issue by moving the set helper module into the package dependencies.

It also upgrades from the deprecated `ember-simple-set-helper` to `ember-set-helper`, and includes the required refactoring.

All tests are passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "ember-concurrency": "^2.1.0",
     "ember-element-helper": "^0.5.5",
     "ember-focus-trap": "^0.7.0",
+    "ember-set-helper": "^2.0.1",
     "ember-truth-helpers": "^3.0.0",
     "tracked-maps-and-sets": "^3.0.1"
   },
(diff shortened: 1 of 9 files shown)


alexlafroscia (approved):
This is awesome; thanks for making this PR!

I had noticed over the weekend that this package was using `ember-simple-set-helper` but didn't realize that it was used inside the `addon` tree and not actually set as a dependency. I really appreciate the clean-up here!

alexlafroscia:
CI doesn't run until approved for first-time contributors, but I just kicked it off. Once it's passing I'll merge and cut a release with this fix!

PR #81, opened 2021-09-08
🐛 `<Listbox>`: 4 bug fixes for scrolling & focus

dmcnamara-eng:
# ✨ What Changed & Why

Fixes for the following 4 `<Listbox>` bugs related to scrolling & focus:

1. 🐛 Prevent arrow keys from scrolling the page https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/8f83502c853be77bf03e0d2e3627fd2f9919c0db
2. 🐛 Scroll selected option into view https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/ebccd5813611dd5ccc0ab6e1a200c0c0cb15b0fa
3. 🐛 Scroll searched option into view https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/855f9e668ccd4a0605dd09c51b06a29d17a84452
4. 🐛 it should be possible to open a listbox without submitting the form https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/d14513ad4c709ed75541667a217006bb00007bbe

### DIFF
--- a/addon/components/listbox.js
+++ b/addon/components/listbox.js
@@ -313,13 +313,14 @@ export default class ListboxComponent extends Component {
     this.search += key.toLowerCase();
 
     for (let i = 0; i < this.optionElements.length; i++) {
+      let optionElement = this.optionElements[i];
+
       if (
-        !this.optionElements[i].hasAttribute('disabled') &&
-        this.optionElements[i].textContent
-          .trim()
-          .toLowerCase()
-          .startsWith(this.search)
+        !optionElement.hasAttribute('disabled') &&
+        optionElement.textContent.trim().toLowerCase().startsWith(this.search)
       ) {
+        this.scrollIntoView(optionElement);
+
         this.activeOptionIndex = i;
         break;
       }
--- a/addon/components/listbox.js
+++ b/addon/components/listbox.js
@@ -182,6 +182,8 @@ export default class ListboxComponent extends Component {
       if (this.args.value === optionComponent.args.value) {
         this.selectedOptionIndex = this.activeOptionIndex =
           this.optionElements.length - 1;
+
+        this.scrollIntoView(optionElement);
       }
     }
 
@@ -242,6 +244,13 @@ export default class ListboxComponent extends Component {
     } else {
       this.optionsElement.focus();
     }
+  }
+
+  scrollIntoView(optionElement) {
+    optionElement.parentElement.scroll(
+      0,
+      optionElement.offsetTop - optionElement.parentElement.offsetTop
+    );
   }
 
   @action
--- a/addon/components/listbox.js
+++ b/addon/components/listbox.js
@@ -8,6 +8,17 @@ const ACTIVATE_NONE = 0;
 const ACTIVATE_FIRST = 1;
 const ACTIVATE_LAST = 2;
 
+const PREVENTED_KEYDOWN_EVENTS = new Set([
+  'ArrowUp',
+  'ArrowDown',
+  'ArrowLeft',
+  'ArrowRight',
+  'PageUp',
+  'PageDown',
+  'Home',
+  'End',
+]);
+
 export default class ListboxComponent extends Component {
   @tracked activeOptionIndex;
   activateBehaviour = ACTIVATE_NONE;
@@ -76,16 +87,7 @@ export default class ListboxComponent extends Component {
 
   @action
   handleKeyDown(event) {
-    if (
-      [
-        'ArrowUp',
-        'ArrowDown',
-        'ArrowLeft',
-        'ArrowRight',
-        'Home',
-        'End',
-      ].indexOf(event.key) > -1
-    ) {
+    if (PREVENTED_KEYDOWN_EVENTS.has(event.key)) {
       event.preventDefault();
     }
   }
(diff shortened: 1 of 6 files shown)


alexlafroscia:
> (It's not sufficient to preventDefault on these events as we handle them, because page scrolling occurs during the capturing phase of event propagation.)

This is a good catch; I was wondering about this!

Is it possible for _us_ to listen for the events during the capture phase as well to call `.preventDefault` rather than needing to add these additional event listeners?

dmcnamara-eng:
> Is it possible for _us_ to listen for the events during the capture phase as well to call `.preventDefault` rather than needing to add these additional event listeners?

https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/0eca5fa69a96625162c17672721ee3092dbb7c59 has a potentially cleaner solution to this. We can  `preventDefault` during the `keydown` event, which is where the browser listens for it.

addon/components/listbox.js
```diff
@@ -57,6 +57,7 @@ export default class ListboxComponent extends Component {
   @action
   handleButtonClick(e) {
     if (e.button !== 0) return;
+    e.preventDefault();
```
dmcnamara-eng, line 60: This change is not specific to page-scroll, but related: 

This fixes a bug when the listbox is embedded in a `<form>`. This prevents the form from being submitted when the trigger button is clicked.

dmcnamara-eng:
@alexlafroscia @GavinJoyce I reckon this is ready to merge. This fixes two important bugs when embedding the listbox in a (long) form.

alexlafroscia:
Hey @dmcnamara-eng, sorry I can't push this through right now, I'm on my honeymoon and am not doing OSS while on vacation. I promise I'll check this out when I return next week! I've been a little neglectful of my OSS duties with the wedding but will be back with energy once I return home.

dmcnamara-eng:
I've cleaned this PR up and commit history for the following 4 `<Listbox>` bugs related to scrolling & focus:

1. 🐛 Prevent arrow keys from scrolling the page https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/8f83502c853be77bf03e0d2e3627fd2f9919c0db
2. 🐛 Scroll selected option into view https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/ebccd5813611dd5ccc0ab6e1a200c0c0cb15b0fa
3. 🐛 Scroll searched option into view https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/855f9e668ccd4a0605dd09c51b06a29d17a84452
4. 🐛 it should be possible to open a listbox without submitting the form https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/d14513ad4c709ed75541667a217006bb00007bbe

We are using this branch in production and these fixes feel solid 👍  

Please 👀 @alexlafroscia 💥

addon/components/listbox.js
```diff
@@ -74,6 +74,22 @@ export default class ListboxComponent extends Component {
...
+  handleKeyDown(event) {
+    if (
+      [
+        'ArrowUp',
+        'ArrowDown',
+        'ArrowLeft',
+        'ArrowRight',
+        'Home',
+        'End',
+      ].indexOf(event.key) > -1
```
alexlafroscia, line 87: Two tiny thoughts (neither of which are blockers here, by any means)

1. We could move this array up to the module scope to avoid creating it on every `keydown` event from the user.
2. We could make it a `Set` instead of an array, which might make the code read a little more easily

```
if (HANDLED_KEYDOWN_EVENTS.has(event.key)) {
  ...
}
```

addon/components/listbox.js
```diff
@@ -223,6 +246,13 @@ export default class ListboxComponent extends Component {
     }
   }
 
+  scrollIntoView(optionElement) {
+    optionElement.parentElement.scroll(
+      0,
+      optionElement.offsetTop - optionElement.parentElement.offsetTop
+    );
```
alexlafroscia, line 253: It seems like `optionElement.scrollIntoView()` has enough browser support that we could call that and simplify this code a bit

https://caniuse.com/?search=scrollIntoView

Thoughts?

alexlafroscia (approved):

addon/components/listbox.js
dmcnamara-eng: 👍 Agreed. Much cleaner and more performant. I've fixed that here: https://github.com/GavinJoyce/ember-headlessui/pull/81/commits/6f34df60eb080085546832ac1fbb291e867a28fa

addon/components/listbox.js
dmcnamara-eng: Aha! `scrollIntoView` was my first go-to, but it ultimately let me down. I've just popped this clarifying comment in the handler:

>    // Cannot use optionElement.scrollIntoView() here because that function
>    // also scrolls the *window* by some amount. Here, we don't want to
>    // jerk the window, we just want to make the the option element visible
>    // inside its container.

alexlafroscia: That's a great caveat and a great comment! Thanks for adding that; I could absolutely see myself trying to refactor this sometime to use `.scrollIntoView` and getting trolled by the difference there.

alexlafroscia (approved):

alexlafroscia:
Thank you so much for all of your patience with me @dmcnamara-eng -- this looks great! Thank you for this PR 🙏

PR #82, opened 2021-09-08
🐛 Listbox: select next/previous option when opening a listbox with no value

dmcnamara-eng:
# ✨ What Changed & Why

Problem: In the case when a listbox has no default `@value`, opening the listbox and navigating with `ArrowUp`/`ArrowDown` does not function.

Solution: Fix `setNextOptionActive` and `setPreviousOptionActive` methods to work even when there is no active option.

# 👹 Before

before

# 😇 After

after

# 🌈 Testing

1. Check out this branch
2. Change `tests/dummy/app/components/listboxes/basic.hbs:3` to have no default `@value`
2. Run `ember serve`
3. Navigate to http://localhost:4200/listbox/listbox-basic
4. Observe `ArrowUp`/`ArrowDown` opens the listbox and navigates up and down the list.


alexlafroscia:
This is a great fix! Could we add a test case to verify the new, correct behavior so that we don't end up with a regression down the line?

PR #84, opened 2021-09-16
🐛  <Transition> pass appear param to child transition

far-fetched:
### What changed and why.
I wanted to update transition documentation for dialogs, and I find out that even though enter classes for child transition are passed correctly, transition is not fired when `appear` parameter is true (initial transition). 

Test checks that initial render is executed when flag is passed. I assume that API for `child` is the same as `<transtion>`, and `appear` was overlooked rather than omitted intentionally.

@alexlafroscia as transition is implemented by you, would be wonderful if you check that fix is fine

### DIFF
--- a/addon/components/transition/child.hbs
+++ b/addon/components/transition/child.hbs
@@ -1,4 +1,5 @@
 <Transition
+  @appear={{@appear}}
   @show={{@show}}
   @unmount={{@unmount}}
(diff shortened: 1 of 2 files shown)


alexlafroscia:
I think you're right @far-fetched -- this should be passed down. Good catch!

PR #85, opened 2021-10-01
Menu: preventing default key event behaviours + adding menu tests

### LINKED ISSUE #80: Preventing default key event behaviours for menu
Hi all, thanks for this add-on. It's been great to see this develop and to learn some of the newer ways of developing with Ember.

Just a quick question on the keyboard events for `<Menu>`. Is there a rationale for not preventing the default behaviour for the up and down arrow events? They are causing some jerky scrolling, when the containing element has a height greater than the viewport. 

The following additions resolve it for me, but I'm unclear if this creates new issues in terms of accessibility? Can't see anything in the w3 docs to suggest it would.
```
  @action
  onKeydown(event) {
    switch (event.key) {
      // Ref: https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-12
      case 'Enter':
        if (this.args.activeItem) {
          this.args.activeItem.element.click();
        }
        this.args.closeMenu();
        break;
      case 'ArrowDown':
+     event.preventDefault();
        return this.args.goToNextItem();
      case 'ArrowUp':
+     event.preventDefault();
        return this.args.goToPreviousItem();
      default:
        if (event.key.length === 1) {
          return this.args.search(event.key);
        }
    }
  }
```
Would you accept a PR to handle these events?

roomman:
This PR is very simple, calling the `preventDefault` and `stopPropagation` methods for key down events in the menu component. This prevents the default scrolling events when up and down arrows are used.

Fixes #80


GavinJoyce:
Thanks for this @roomman. Perhaps you would add a test to ensure that we don't regress here in future?

GavinJoyce:
Looking great @roomman, nice one. There are some commented out lines with JSX in them, perhaps you could remove them and squash into a single commit when you're ready to merge?

roomman:
Sorry @GavinJoyce that's supposed to be wip, committed to my local branch. My bad. Will submit the full PR when done 👍🏻

GavinJoyce:
👍 sounds good, thanks

roomman:
@GavinJoyce here's the full PR as promised. I'll circle back and add tests for the remaining keydown modules in the next week or so, subject to workload.

Originally, I was supposed to be addressing your comment above, about protecting against regressing for `event.preventDefault` on the up and down events. Without adding something like sinon to facilitate spying on the event methods, I'm not sure how to approach this. Any thoughts - happy to write the test if you can signpost. 👍🏻

GavinJoyce:
Just poking about in the dummy app, these changes seem to introduce a bug where pressing `Enter` on an active items doesn't close the menu. Pressing enter a second time throws an exception:

1. Open the menu

Screenshot 2021-10-04 at 17 50 45

2. Navigate to an item and press `Enter` - the menu stays open:

Screenshot 2021-10-04 at 17 50 53

3. Press `Enter` a second time, an exception is thrown:
4. 
Screenshot 2021-10-04 at 17 50 59

roomman:
Hi @GavinJoyce, so it turns out that the bug is unrelated to this PR, and was introduced [in this commit](https://github.com/GavinJoyce/ember-headlessui/commit/e46741d43ef94f903e812ee67b2e5dcd9bfc66f6).

My best guess is that it's related to the introduction of the `{{focus-trap}}` modifier, and appears to have resolved itself thanks to the work done by @far-fetched in #98. I've merged master into this PR and the bug has gone.

Are you happy for me to add some additional tests to the `ArrowDown` and `ArrowUp` modules to catch this scenario, despite them being additional to headless-ui test suite?

roomman:
That said, I just wrote a test, on a branch where the bug is present, navigating to a menu item using `ArrowDown` and invoking with `Enter` and the test passes. 🤦🏻‍♂️

GavinJoyce:
Thanks @roomman, this is great!

PR #87, opened 2021-10-01
Add PopperJS primitive

NullVoxPopuli:
- [ ] Needs TypeScript
- [ ] Needs `@popperjs/core`
- [ ] Needs a function-modifier implementation
- [ ] Requires at least Ember 3.25

Copied from [here](https://github.com/NullVoxPopuli/limber/blob/56850424ec79dba743d783917c423291e094b973/frontend/app/components/popper-j-s.ts)
Example usage [here](https://github.com/NullVoxPopuli/limber/blob/56850424ec79dba743d783917c423291e094b973/frontend/app/components/limber/menu/index.hbs#L2)

### DIFF
--- a/addon/components/popper-j-s-.ts
+++ b/addon/components/popper-j-s-.ts
@@ -1,0 +1,90 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { setComponentTemplate } from '@ember/component';
+import { registerDestructor } from '@ember/destroyable';
+import { action } from '@ember/object';
+import { hbs } from 'ember-cli-htmlbars';
+
+import { createPopper } from '@popperjs/core';
+// could use ember-modifier here as well
+import { modifier } from 'ember-could-get-used-to-this';
+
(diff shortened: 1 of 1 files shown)


GavinJoyce:
Does `headlessui` have a popper primitive? I seem to remember that they used it in their examples alright, but I don't recall it being part of the library.

Here's one place in the docs where they mention optionally using popper

NullVoxPopuli:
I mean, I can make a separate ember popper library, which is what they done for react to get those hooks.
It's impossible to use pepper directly without a bunch of boilerplate

GavinJoyce:
Sorry, I wan't clear. 

Does `headlessui` include popper as a dependency? AFIAK, they don't - they just give an example of how to use it in their samples.

NullVoxPopuli:
ah yes! I got you now -- yes that's what I observed as well.

I added this PR because I saw we wanted a Menu + Popper demo.
So, we'll do the same? _just_ include popper in the demos?

GavinJoyce:
Sounds great, thanks

PR #89, opened 2021-10-01
Add typescript

NullVoxPopuli:
Resolves: https://github.com/GavinJoyce/ember-headlessui/issues/18

- [x] Requires support for linting typescript
- [x] Convert at least one thing to typescript

PR #90, opened 2021-10-01
Prepare for typescript

NullVoxPopuli:
- Unblocks: https://github.com/GavinJoyce/ember-headlessui/pull/89
  - Which Unblocks: https://github.com/GavinJoyce/ember-headlessui/pull/87
    - Which Unblocks: adding a Menu with PopperJS
    
    
For review:
 - you'll want to review per-commit
 - these changes are mostly superficial -- no behavioral changes expected
 
 regarding the lint changes... configuring linting has gotten too hard with all the different projects variants out there... so I have a meta lint package that makes it easy.. let me know if you don't agree with it


.eslintrc.js
```diff
@@ -1,57 +1,19 @@
...
-      ],
-      parserOptions: {
-        sourceType: 'script',
-      },
-      env: {
-        browser: false,
-        node: true,
+      files: ['**/*.js', '**/*.ts'],
+      rules: {
+        'prettier/prettier': ['error', require('./.prettierrc')],
```
NullVoxPopuli, line 15: the defaults from the lint meta-package have the printWidth set to 100, and that caused too many changes, so I figured it would be best to just use the existing prettier config

GavinJoyce (approved):
nice one!

.eslintrc.js
```diff
@@ -1,57 +1,19 @@
 'use strict';
 
+const { configs } = require('@nullvoxpopuli/eslint-configs');
```
GavinJoyce, line 3: perhaps it would simpler to keep the config boilerplate here and avoid the indirection?

NullVoxPopuli: I don't simpler, but def clearer as to what's happening. Linting in hybrid projects unfortunately isn't all that simple anymore :(

If you're curious though, here is the config:
https://github.com/NullVoxPopuli/eslint-configs/blob/main/configs/ember.js#L141

GavinJoyce: 👍 thanks. Agree it's not simpler, but perhaps clearer.

It'll only become an issue if we need to modify this at some point, that may never happen though. Happy to ship this as is and revisit if that ever happens

NullVoxPopuli: So _far_, modifying has been easy, as demonstrated here: https://github.com/NullVoxPopuli/limber/blob/main/frontend/.eslintrc.js#L14 (which, this specific modification is now obsolete). 

But yeah, I hear you. In this eslint-configs package, I should probably add a `bin` to "eject" the config, kinda like create-react-app does for all the webpack configs 🤔

Issue #77, opened 2021-09-01
Use `@as` to customize root element


alexlafroscia:
Some of our component use a `@tagName` argument to customize the root element, while others use `@as`

We should use `@as` across _all_ components for two main reasons:

1. Consistency with the React/Vue implementations 
2. Accepting a component rather than _just_ a string would be nice, like `Menu::ItemElement` does (to support `LinkTo`)
    https://github.com/GavinJoyce/ember-headlessui/blob/e936dff7d953e293e5b05bf69459d421183ebba5/addon/components/menu/item-element.hbs#L1-L4

Note: this will be a breaking change for the components currently configured to use `@tagName`

NullVoxPopuli:
I'm in favor

GavinJoyce:
Might be nice to create a helper named something like `componentOrElement` to remove the need for a backing class in the `Menu::ItemElement` component:

```hbs
{{#let (componentOrElement @as) as |Tag|}}

```

GavinJoyce:
Hmm, ^ might not be possible as the `element` helper is actually an AST transform, not a function that we can invoke:

https://github.com/tildeio/ember-element-helper/blob/main/addon/helpers/element.js

NullVoxPopuli:
do we have an RFC for an element helper? maybe we need to implement that over in the glimmer-vm?

GavinJoyce:
A quick improvement might be to just create a `tagNameIsComponent` helper, that will allow us to remove the need for a backing class

GavinJoyce:
> do we have an RFC for an element helper?

The https://github.com/tildeio/ember-element-helper readme references a couple of RFCs

bertdeblock:
👋 Interested in contributing to this project and this seems like a good one to pick up (unless someone is already working on this of course).

Couple of questions:
- Should a tag name and a component _always_ be accepted everywhere or are there cases were we only want to allow a tag name and not a component or vice versa?
- Should the current `@as` cases be updated to also accept components?
- Should `@tagName` be deprecated first or do we just delete it and release a new minor?
- It seems `{{tag-name-is-component}}` does not support imported component classes (e.g. `@as={{this.SomeGlimmerComponentClass}}`) would it be okay to update this? Can be a separate PR of course

PR #91, opened 2021-10-03
create a `tag-name-is-component` helper

GavinJoyce:
part of https://github.com/GavinJoyce/ember-headlessui/issues/77#issuecomment-932958600

This removes the need for the `Menu::ItemElement` component to have a backing class.

An existing test is covering this: https://github.com/GavinJoyce/ember-headlessui/commit/83796b38b6829e62c631c94e56c0542a6828156d#diff-2f1746c8a31c717bfe0069265e1afc57117088ae1761e4a498ecaa6ebc325897R157

### DIFF
--- a/addon/helpers/tag-name-is-component.js
+++ b/addon/helpers/tag-name-is-component.js
@@ -1,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+function tagNameIsComponent([as]) {
+  return typeof as === 'object';
+}
+
+export default helper(tagNameIsComponent);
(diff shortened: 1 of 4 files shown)


addon/helpers/tag-name-is-component.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+function tagNameIsComponent([as]) {
+  return typeof as === 'object';
```
NullVoxPopuli, line 4: do we want this to return true for things that aren't components?

GavinJoyce: Technically no, but it would be weird for someone to pass an object that isn't a component.

Perhaps there are some other things we could check for on the object? If not, I think this is likely good enough for now.

NullVoxPopuli: I mean, this is a quick win, we can make it more constrained later as the public API declares what is intended anyway

NullVoxPopuli (approved):

Issue #88, opened 2021-10-01
Ditch @ember/render-modifiers


NullVoxPopuli:
We should provide our own modifier implementation for now, because `@ember/render-modifiers` throws a deprecation in Ember 3.22+ _and_ render-modifiers are anti-patternyish https://nullvoxpopuli.com/avoiding-lifecycle

NullVoxPopuli:
Potentially, we could go this approach so that we are compatible with Ember 4 once the old modifier managers are removed: https://github.com/NullVoxPopuli/ember-popperjs/blob/main/addon/-private/-function-modifier.ts

NullVoxPopuli:
Resolved in https://github.com/GavinJoyce/ember-headlessui/pull/154

PR #92, opened 2021-10-03
Ditch render-modifiers

NullVoxPopuli:
WIP - exploration for now

Begins #88

### DIFF
--- a/addon/utils/function-modifier.ts
+++ b/addon/utils/function-modifier.ts
@@ -1,0 +1,84 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import {
+  capabilities as modifierCapabilities,
+  setModifierManager,
+} from '@ember/modifier';
+
+type LazyTrackedArgs = {
+  positional?: Array<unknown>;
+  named?: Record<string, unknown>;
+};
+
+type FunctionModifier<Args extends LazyTrackedArgs> =
+  | ((
+      element: HTMLElement,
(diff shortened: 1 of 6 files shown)


addon/utils/function-modifier.ts
```diff
@@ -0,0 +1,84 @@
...
+  element?: HTMLElement;
+  destructor?: void | (() => void);
+}
+
+/**
+ *  This is copied from
+ *  https://github.com/pzuraq/ember-could-get-used-to-this/blob/master/addon/-private/modifiers.js
+ *
+ * Because:
+ *  - we don't want to *force* folks into use ember-could-get-used-to-this
```
GavinJoyce, line 34: No against including this here, just wondering why we don't want to depend on `ember-could-get-used-to-this`?

NullVoxPopuli: we absolutely could! what other things might we want to use `ember-could-get-used-to-this` for?
(I think we need to move ember-could-get-used-to-this in to adopted-addons, or specifically extract out the function-modifier bit to a new addon and have that included everywhere -- maybe call it `ember-default-modifier-manager-polyfill` or something related to the matching RFC 🤷)

NullVoxPopuli:
superseded by https://github.com/GavinJoyce/ember-headlessui/pull/116

NullVoxPopuli:
superseded by: https://github.com/GavinJoyce/ember-headlessui/pull/116

Issue #34, opened 2021-05-21
Enable embroider ember-try scenarios


GavinJoyce:
I disabled them in https://github.com/GavinJoyce/ember-headlessui/pull/33 as they were causing github actions to timeout

PR #33, opened 2021-05-21
disable EMBER_TRY_SCENARIO=embroider-optimized to fix build

GavinJoyce:
trying to address https://github.com/GavinJoyce/ember-headlessui/pull/30#issuecomment-845713887

PR #93, opened 2021-10-03
Re-enable embroider testing #34, #33, #72

NullVoxPopuli:
Supersedes #72


GavinJoyce (approved):
💯

NullVoxPopuli:
ah, we have to resolve this:
```
- Global error: Uncaught ReferenceError: Ember is not defined at 
```

NullVoxPopuli:
looks like we have to remove ember-test-selectors: 
image

(or PR to them and block _this_ PR)

NullVoxPopuli:
@GavinJoyce + @alexlafroscia + @achambers do ya'll have thoughts on _leaving_ the test selectors in, and then providing a (fractal) page object in addon-test-support for enabling easier interaction with these components in consuming apps? so that way consumers don't all need to write their own test helpers. (or they could, and just ignore what is provided, like if they don't like fractal for whatever reason (I only mention fractal, because ember-cli-page-object does _not_ work well with typescript, and is generally _too_ dynamic)). Another option is go page-object agnostic, and "just use the js apis" (but then we don't have any integrated query selector nesting, which is mostly what the page object patterns are for anyway (I mean, secondary to providing a humanized API for component interaction, and abstracting away the DOM)

GavinJoyce:
I'm all for providing useful test helpers to consuming apps. I've no experience with fractal page objects, but happy to use them if they provide consuming apps a nice test API

tests/index.html
```diff
@@ -16,6 +16,13 @@
 
     {{content-for "head-footer"}}
     {{content-for "test-head-footer"}}
+
+    <script>
+      // testdouble support
+      // testdouble doesn't natively support the browser as it assumes
+      // that a `global` identifier will be available, but that is not the case for browsers
```
NullVoxPopuli, line 23: I reported this here: https://github.com/testdouble/testdouble.js/issues/438
and it seemed there was 0 interest in resolving... :-\

NullVoxPopuli: I also opened a new issue: https://github.com/testdouble/testdouble.js/issues/471

to see what they feel like doing about it 🤷

NullVoxPopuli:
alright, next blocker: https://github.com/mixonic/ember-cli-deprecation-workflow/issues/133

alexlafroscia:
My perspective on the test helpers:

- I would like to see the `addon` tree _not_ ship any `data-test-` selectors; I feel like this is a testing detail that should be private to our own tests, or configured by the app using Headless UI, but isn't something we should push onto our users. They become part of the public API if we include them, and I would rather we avoid that altogether. That would also allow us to avoid needing `ember-test-selectors` at all as a dependency
- I would like to see the addon ship the test selectors we are using inside the `addon-test-support` directory so that host apps can test their components the same way they are tested internally. Whether they are functions or page objects doesn't matter to me, though I like the idea that it's just functions as it avoids being tied to any given page object implementation

PR #95, opened 2021-10-03
Add menu with popper example

NullVoxPopuli:
This is a breaking change, so if ya'll don't want to go along with this, that's fine by me.

the breaking change is that any consuming project must also be using ember-auto-import@v2

### DIFF
--- a/tests/dummy/app/templates/menu/menu-with-popper.hbs
+++ b/tests/dummy/app/templates/menu/menu-with-popper.hbs
@@ -1,1 +1,69 @@
-menu-with-popper will go here
+<div class='flex justify-center w-screen h-full p-12 bg-gray-50'>
+  <div class='relative inline-block text-left'>
+    <Menu as |menu|>
+      <PopperJS as |trigger popover|>
+        <span class='rounded-md shadow-sm'>
+          <menu.Button
+            {{trigger}}
+            class='inline-flex justify-center w-full px-4 py-2 text-sm font-medium leading-5 text-gray-700 transition duration-150 ease-in-out bg-white border border-gray-300 rounded-md hover:text-gray-500 focus:outline-none focus:border-blue-300 focus:shadow-outline-blue active:bg-gray-50 active:text-gray-800'
+          >
+            <span>Options</span>
+            <svg
+              class='w-5 h-5 ml-2 -mr-1'
+              viewBox='0 0 20 20'
+              fill='currentColor'
+            >
+              <path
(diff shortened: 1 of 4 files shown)


tests/dummy/app/templates/menu/menu-with-popper.hbs
```diff
@@ -1 +1,69 @@
-menu-with-popper will go here
\ No newline at end of file
+<div class='flex justify-center w-screen h-full p-12 bg-gray-50'>
```
NullVoxPopuli, line 1: this is menu/basic, but with PopperJS

GavinJoyce:
> the breaking change is that any consuming project must also be using ember-auto-import@v2

I think it's fine for us to make breaking changes like this 👍 

FYI, it looks like there are some failing tests in CI

NullVoxPopuli:
excellent! because ember-auto-import@v2 is a requirement for ember 4 :D

NullVoxPopuli:
I'm not sure how any of these were passing 🤔 
image

I assume most of this was copied from either the React or Vue repo?

NullVoxPopuli:
It seems back when we were on webpack 4 (ember-auto-import@v1, a bunch of implicit non-browser behaviors were getting polyfilled)

tests/accessibility-assertions.js
```diff
@@ -129,86 +129,81 @@ function assertDialog(options, dialog = getDialog()) {
...
 
-  try {
-    switch (options.state) {
-      case DialogState.InvisibleHidden: {
-        if (dialog === null) {
-          Qunit.assert.notEqual(dialog, null);
-
-          return;
-        }
+  switch (options.state) {
```
NullVoxPopuli, line 132: best to view this file without invisible character changes

tests/accessibility-assertions.js
```diff
@@ -129,86 +129,81 @@ function assertDialog(options, dialog = getDialog()) {
...
+    case DialogState.InvisibleUnmounted: {
+      Qunit.assert.equal(dialog, null);
+      break;
+    }
+
+    default: {
+      assertNever(options.state);
     }
-  } catch (err) {
-    Error.captureStackTrace(err, assertDialog);
```
NullVoxPopuli, line 210: captureStackTrace is a node-only API

GavinJoyce:
FYI, I've updated `ember-auto-import` here: https://github.com/GavinJoyce/ember-headlessui/pull/111

PR #98, opened 2021-10-04
➖ Remove ember focus trap addon

far-fetched:
We still have some skipped / todo tests in dialog, listbox, menu(?) which are related to focused element after close component or click outside. 
[dialog test](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/dialog-test.js#L411)
[listbox](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/listbox-test.js#L2985)

To sort them out, I added some useful functionality to `focus-trap` package: [link](https://github.com/focus-trap/focus-trap/pull/497)
Last step would be to update `ember-focus-trap` with new version of `focus-trap`. When I tried to do it locally and fix remaining tests, I figured out that `ember-focus-trap` adds some extra functionality to it. Currently, we don't use any of them + one of the functionality causes a bug, which I described [here](https://github.com/josemarluedke/ember-focus-trap/issues/40). 
TL TR: `ember-focus-trap` replicates behavior which is already implemented in `focus-trap`

With this issue I think we are not able to fix those remaining tests. This PR removes `ember-focus-trap` in favor of own bare minimum implementation. Tests are still passing, so I assume that I didn't change any behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,10 +40,11 @@
     "ember-click-outside-modifier": "^2.0.0",
     "ember-concurrency": "^2.1.0",
     "ember-element-helper": "^0.5.5",
-    "ember-focus-trap": "^0.7.0",
     "ember-set-helper": "^2.0.1",
     "ember-truth-helpers": "^3.0.0",
-    "tracked-maps-and-sets": "^3.0.1"
+    "focus-trap": "^6.7.1",
+    "tracked-maps-and-sets": "^3.0.1",
+    "ember-auto-import": "^1.11.2"
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -40,10 +40,11 @@
     "ember-click-outside-modifier": "^2.0.0",
     "ember-concurrency": "^2.1.0",
     "ember-element-helper": "^0.5.5",
-    "ember-focus-trap": "^0.7.0",
     "ember-set-helper": "^2.0.1",
     "ember-truth-helpers": "^3.0.0",
-    "tracked-maps-and-sets": "^3.0.1"
+    "focus-trap": "^6.7.1",
+    "tracked-maps-and-sets": "^3.0.1",
+    "ember-auto-import": "^1.11.2"
```
far-fetched, line 47: I moved this dep (`ember-auto-import`) here, because `focus-trap` is not imported correctly when `ember-auto-import` was in dev.

GavinJoyce:
Thanks, is the plan to re-enable the skipped tests?

far-fetched:
Yep I can do this, just after we merge this PR which updates `focus-trap` dep with new functionality.

GavinJoyce:
👍 sounds great, thanks

PR #108, opened 2021-11-01
✨ <Dialog/> Set return focus on clicked element

far-fetched:
### **What changed and why**

PR adds `setReturnFocus` option from `focus-trap` where we can explicitly set which element should be focus after deactivation. Default behavior is to return focus to the element which opened the dialog (trigger button). 
When there is no overlay, and the user clicks outside the dialog, focus will be set to the element which is clicked by the user (see test).
By the way of implementing it, I figure out that `focus-trap` also uses similar pattern to detect `outside-click`, so far we used dedicated modifier (`outside-click-modifier`) as we use it only here I think we can use `focus-trap` and safely remove `outside-click-modifier` from dependency list.

### DIFF
--- a/addon/components/dialog.hbs
+++ b/addon/components/dialog.hbs
@@ -7,9 +7,12 @@
         aria-modal='true'
         ...attributes
         {{headlessui-focus-trap
-          focusTrapOptions=(hash initialFocus=@initialFocus)
+          focusTrapOptions=(hash
+            initialFocus=@initialFocus
+            allowOutsideClick=this.allowOutsideClick
+            setReturnFocus=this.setReturnFocus
+          )
         }}
-        {{click-outside this.onClose event='mouseup'}}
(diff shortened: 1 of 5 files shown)


GavinJoyce:
Thanks for this. Does this mirror the behaviour in React and Vue?

far-fetched:
As react and vue use their own implementation of `focus-trap` [link](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/focus-trap/focus-trap.tsx) with hooks - implementation is different, but final behavior stays the same - [test in react](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/dialog/dialog.test.tsx#L625-L658) is almost the same (semantic difference).

GavinJoyce:
Great, thanks!

PR #109, opened 2021-11-03
✅ update dialog test - composition with transitions

far-fetched:
As we already have `<Transition/>` component, we can update todo test from dialog. Highly inspired from [react repo ](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/dialog/dialog.test.tsx#L356)


GavinJoyce:
Nice, thanks

PR #110, opened 2021-11-12
fix(dialog): lock scroll when the dialog is open

alexlafroscia:
While working with the `Dialog` component, @JessWallin and I noticed that we're not locking the scroll of the page behind the `Dialog` when it's open. This is something we should be doing:

- The documentation for the "real" HeadlessUI [mentions it](https://headlessui.dev/react/dialog#other)
- The code for how the React implementation does this is [here](https://github.com/tailwindlabs/headlessui/blob/e9e6aded545c329585d8cf7452ad9fe400a5406b/packages/%40headlessui-react/src/components/dialog/dialog.tsx#L229-L246)

This PR adds a modifier to `Dialog` that follows the exact same approach as the React version to locking the page scroll.

The _one_ difference is that the React implementation applies the styles to `document.documentElement`, which is the `html` element, while we decided to use `$portalRoot` (which is likely `document.body` instead). This seems to work just as well, and we looked around online for other instances of scroll-locking; many of them style `body` rather than `html`. Using `$portalRoot` means that only the test preview window is locked, which seemed like a helpful way for us to avoid messing with the QUnit UI during tests that make use of `Dialog`!

PR co-authored by @JessWallin

### DIFF
--- a/addon/components/dialog.ts
+++ b/addon/components/dialog.ts
@@ -10,7 +10,14 @@ import { modifier } from 'ember-modifier';
 
 import type DialogStackProvider from 'ember-headlessui/services/dialog-stack-provider';
 
-function getPortalRoot() {
+/**
+ * Expose the element that the `Dialog` should be "slotted" into
+ *
+ * This is exported _only_ for testing purposes; do not consider this API to be public
+ *
+ * @private
+ */
+export function getPortalRoot() {
   const { rootElement } = getOwnConfig();
 
   // If we looked up a `rootElement` config at build-time, use that; otherwise use the body
@@ -51,6 +58,29 @@ export default class DialogComponent extends Component<Args> {
       };
     }
   );
+
+  lockWindowScroll = modifier(() => {
+    // Opt-out of some other dialog already locked scrolling
+    if (this.dialogStackProvider.dialogIsOpen) {
+      return;
+    }
+
+    let overflow = this.$portalRoot.style.overflow;
+    let paddingRight = this.$portalRoot.style.paddingRight;
+
+    // Setting `overflow: hidden` will suddenly hide the scroll bar on the window, which can cause horizontal
+    // layout shifting when the `Dialog` becomes open
+    // By applying the width of the scroll bar as padding, we can avoid that layout shift from happening
+    let scrollbarWidth = window.innerWidth - this.$portalRoot.clientWidth;
+
+    this.$portalRoot.style.overflow = 'hidden';
+    this.$portalRoot.style.paddingRight = `${scrollbarWidth}px`;
+
+    return () => {
+      this.$portalRoot.style.overflow = overflow;
+      this.$portalRoot.style.paddingRight = paddingRight;
+    };
+  });
 
   constructor(owner: unknown, args: Args) {
     super(owner, args);
--- a/tests/integration/components/dialog-test.js
+++ b/tests/integration/components/dialog-test.js
@@ -9,6 +9,7 @@ import { module, test, todo } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 
 import userEvent from '@testing-library/user-event';
+import { getPortalRoot } from 'ember-headlessui/components/dialog';
 import { Keys } from 'ember-headlessui/utils/keyboard';
 
 import {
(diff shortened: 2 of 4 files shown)


addon/components/dialog.ts
```diff
@@ -10,7 +10,7 @@ import { modifier } from 'ember-modifier';
 
 import type DialogStackProvider from 'ember-headlessui/services/dialog-stack-provider';
 
-function getPortalRoot() {
+export function getPortalRoot() {
```
alexlafroscia, line 13: This function has been exported so that the test code can target this element in the exact same way. Sharing the "lib" code with the tests means that we can change them both together if the root element changes somewhere down the road.

addon/components/dialog.ts
```diff
@@ -52,6 +52,26 @@ export default class DialogComponent extends Component<Args> {
     }
   );
 
+  lockWindowScroll = modifier(() => {
+    // Opt-out of some other dialog already locked scrolling
+    if (this.dialogStackProvider.dialogIsOpen) {
```
alexlafroscia, line 57: We added this `dialogIsOpen` property to the service because there really wasn't an API for querying whether a dialog is open generically. The `hasOpenChild` method isn't really expressing what we want here; we want to know if this modal, being opened right now, is the "top most" modal. If it is, we want to run the scroll lock. If not, we will rely on the "top most" modal already having applied the lock, and trust they will clean it up as well.

tests/integration/components/dialog-test.js
```diff
@@ -233,7 +234,50 @@ module('Integration | Component | <Dialog>', function (hooks) {
...
+          );
+
+        await click('[data-test-overlay]');
+
+        assert
+          .dom(portalRoot)
+          .doesNotHaveStyle(
+            { overflow: 'hidden' },
+            'The page is "unlocked" when the dialog closes'
+          );
```
alexlafroscia, line 279: Technically this assertion is "extra"; the upstream tests for the React implementation don't test cleaning things up after the `Dialog` is closed, but that seemed like a useful thing for us to have!

addon/components/dialog.ts
NullVoxPopuli: should we mark all exported functions not intended for public consumption with `// @private`?

addon/components/dialog.ts
```diff
@@ -52,6 +52,26 @@ export default class DialogComponent extends Component<Args> {
...
+      return;
+    }
+
+    let overflow = this.$portalRoot.style.overflow;
+    let paddingRight = this.$portalRoot.style.paddingRight;
+
+    let scrollbarWidth = window.innerWidth - this.$portalRoot.clientWidth;
```
NullVoxPopuli, line 64: can we add a comment here about how this is to prevent layout jitter when the scrollback is hidden due to overflow: hidden being set.
You and I both know why this is here, but I'd like stuff like this to be documented in code to lower the barrier to entry and to enhance understand for broader contributions, ya know? <3

NullVoxPopuli (approved):
LGTM! 🎉

addon/components/dialog.ts
alexlafroscia: Yeah, that's a great idea! Good thinking.

addon/components/dialog.ts
alexlafroscia: Great idea! I'll do that 👍

alexlafroscia:
Rebased to add those recommended comments -- thanks @NullVoxPopuli!

addon/components/dialog.ts
NullVoxPopuli: 🙌

NullVoxPopuli (approved):

tests/integration/components/dialog-test.js
GavinJoyce: I'm not sure why, but this test seems to fail when I run it locally:

Screenshot 2021-11-20 at 08 13 00

Does it pass for you?

alexlafroscia: Yup, it's passing locally for me

For some reason a bunch of the `Listbox` tests fail, but the `Dialog` ones all pass

GavinJoyce: Weird, thanks. I'll try on another machine

PR #111, opened 2021-11-20
upgrade ember-auto-import

GavinJoyce:
extracted from https://github.com/GavinJoyce/ember-headlessui/pull/95

So as to not block getting this addon ready for ember 4, I'm temporarily skipping tests that were failing with `t is not a function` errors https://github.com/GavinJoyce/ember-headlessui/issues/113. We can reenable them once we've figured our the cause

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -17,6 +17,7 @@ module.exports = async function () {
       },
       {
         name: 'ember-release',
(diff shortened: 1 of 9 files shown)


GavinJoyce:
seems like https://github.com/alexlafroscia/qunit-wait-for is not playing nicely with the `ember-auto-import` upgrade

PR #120, opened 2021-12-18
Move `ember-test-selectors` to `dependencies`

bertdeblock:
This makes sure that `data-test` selectors are also stripped when the consuming app doesn't have `ember-test-selectors` installed itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "ember-concurrency": "^2.1.2",
     "ember-element-helper": "^0.5.5",
     "ember-set-helper": "^2.0.1",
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Personally, I don't want addons to cause extra build work for consuming apps.
I think the best long-term thing to do would be to remove ember-test-selectors either entirely from this package -- or maybe during build time --like as a rollup plugin (would require v2 conversion first tho)

bertdeblock:
Removing `ember-test-selectors` is fine for me as well. That's something that can be done right away actually because as is, it literally does nothing at the moment.

PR #121, opened 2021-12-18
Fix `it should add a scroll lock to the html tag` test

bertdeblock:
I just checked out the project locally and this test was failing consistently for me. I'm a bit confused as to why, because it seems this doesn't happen during CI? It seems the issue is that the dialog overlay is rendered outside of the testing container which is used as the root when querying elements. Using the `getDialogOverlay` helper fixes this as it uses the `document` as the root to get the dialog overlay element. The same is done [here](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/dialog-test.js#L449) and [here](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/dialog-test.js#L530).

### DIFF
--- a/tests/integration/components/dialog-test.js
+++ b/tests/integration/components/dialog-test.js
@@ -248,7 +248,7 @@ module('Integration | Component | <Dialog>', function (hooks) {
             @onClose={{set this "isOpen" false}}
             as |d|
           >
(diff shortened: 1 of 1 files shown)


GavinJoyce:
Thanks @bertdeblock 🍻

Issue #113, opened 2021-11-21
Some tests fail after upgrading `ember-auto-import`


GavinJoyce:
The failures seem to be related to https://github.com/alexlafroscia/qunit-wait-for:

<img width="1019" alt="Screenshot 2021-11-21 at 14 16 34" src="https://user-images.githubusercontent.com/2526/142765461-67d3b02a-1dea-43b0-840e-e2d94519a462.png">

So that we're not blocking to the move to Ember 4.0, I'm going to disable the tests for now. We can either fix the issue and reenable, or rewrite the tests to no use `assert.waitFor`.

GavinJoyce:
@alexlafroscia perhaps you have seen something like this before with `ember-wait-for`?

NullVoxPopuli:
is this still an issue?

GavinJoyce:
I believe so, there are a bunch of tests which are currently being skipped: https://github.com/GavinJoyce/ember-headlessui/search?q=issues%2F113

GavinJoyce:
Thanks for the PRs @bertdeblock!

I think I know why tests sometimes fail locally. If you click in the browser while the tests are running, it's possible to get tests to fail. eg. If you click the "Hide passes tests" while the tests are executing, you might get a failure like this:

<img width="1593" alt="Screenshot 2021-12-20 at 09 27 02" src="https://user-images.githubusercontent.com/2526/146745112-ec43a645-9e8f-4247-8ad5-88d9c2952478.png">

bertdeblock:
Same goes for tests that rely on an element being in focus and the test container is hidden. Happened in this project as well I think. I've created a PR over in `ember-qunit` that should resolve this:
https://github.com/emberjs/ember-qunit/pull/912

NullVoxPopuli:
is this issue still a problem?

PR #122, opened 2021-12-19
Enable skipped dialog tests again

bertdeblock:
Seemed like a good one to pick up. Surprisingly all tests, but one, ran fine for me locally.

I will try to have a look at the other skipped ones as well.

Related to #113.


bertdeblock:
One failing test, but seems unrelated. Will see if I can reproduce locally or does this one fail sometimes during CI? Didn't notice this when running the suite locally.

bertdeblock:
@GavinJoyce Would you mind punching CI again? I can get the test to fail locally as well, but it only happens 1 out of 20 times or so.

GavinJoyce:
sure, they are running again now

PR #123, opened 2021-12-19
Enable skipped transition tests again

bertdeblock:
These seem to work fine locally so let's see what CI says 😄.

Related to #113.


GavinJoyce:
Thanks, these are passing for me locally too

PR #128, opened 2022-01-13
✨ <Menu /> support static option for Items

aoifehannigan:
This PR add support for the `static` option in the  MenuItems component which is used to determine whether the element should ignore the internally managed open/closed state.

Prop: `static`
Type: `Boolean`
Default: `false`

### DIFF
--- a/addon/components/menu/items.hbs
+++ b/addon/components/menu/items.hbs
@@ -1,5 +1,5 @@
 {{! template-lint-disable no-down-event-binding }}
-{{#if @isOpen}}
+{{#if (or @isOpen @static)}}
   <div
     id={{@itemsGuid}}
     aria-labelledby={{@buttonGuid}}
--- a/tests/integration/components/menu-test.js
+++ b/tests/integration/components/menu-test.js
@@ -219,6 +219,21 @@ module('Integration | Component | <Menu>', (hooks) => {
       assert.dom('[data-test-item-a]').hasTagName('a');
       assert.dom('[data-test-item-a]').hasAttribute('href', '/menu');
     });
+
+    test('should be possible to always render the Menu.Items if we provide it a `static` prop', async function (assert) {
+      await render(hbs`
+        <Menu as |menu|>
+          <menu.Button>Trigger</menu.Button>
+          <menu.Items data-test-menu-items @static={{true}} as |items|>
+              <items.Item>Item A</items.Item>
+              <items.Item>Item B</items.Item>
+              <items.Item>Item C</items.Item>
+          </menu.Items>
+        </Menu>
+      `);
+
+      assert.dom('[data-test-menu-items]').isVisible();
+    });
   });
 
   module('Keyboard interactions', function () {


addon/components/menu/items.hbs
```diff
@@ -1,5 +1,5 @@
 {{! template-lint-disable no-down-event-binding }}
-{{#if @isOpen}}
+{{#if (or @isOpen @static)}}
```
NullVoxPopuli, line 2: do we need `@static`? I really don't like over-expanding public API, and it seems just setting `@isOpen={{true}}` would be sufficient?

NullVoxPopuli:
Hello!!! thank you for your contribution :sparkles: :heart: 
Just one question -- which which I can be persuaded for -- I just don't understand the use case atm. :sweat_smile:

addon/components/menu/items.hbs
bertdeblock: It seems the other implementations support a `static` option as well.

aoifehannigan: Yes I included the addition of the `static` option as it is supported in the [React & Vue implementations](https://headlessui.dev/react/menu#menu-items) and is already available in the [ListBoxOptions](https://github.com/GavinJoyce/ember-headlessui/blob/master/addon/components/listbox/-options.hbs) component

In terms of a use case we will be using multiple Menus in a left nav and we need don't require the dropdown to close when focus is lost, only if the user has clicked to close the Menu, or another Menu has been opened. Utilising the `static` option and to override the behavior in our app seemed like the best way to achieve this.

far-fetched: Regarding use case:
Somewhere in react doc, I saw information that might be useful for handling with own transition and animation.

aoifehannigan: You're right it's [here](expandedNav) thank you!

NullVoxPopuli: What's the advantage over using isOpen though? I'm confused. Are the semantics really that much better even though the arg does the exact same thing?
Is there a difference in the other implementation?
🤔

NullVoxPopuli:
I guess the api to menu is already pretty small.
My main concern is eventual api bloat as she many ui libraries tend to have (8+ args, etc)

If we can keep that in check, and diverge from the other implementations if we need to, I'm ok with this change.

I mostly want to protect our users if react or vue libraries go off the rails.

addon/components/menu/items.hbs
far-fetched: hmm I think that `<Menu.Items/>` does not expect to receive `isOpen` prop. It is in template because is passed down from `<Menu/>` component, but intention was to not override this prop.
[Here ](https://headlessui.dev/react/menu#showing-hiding-the-menu)

> By default, your Menu.Items instance will be shown/hidden automatically based on the internal open state tracked within the Menu component itself.

Probably it is possible to do it without `static` arg, but then the question is how far we want to be from original API in react and vue

NullVoxPopuli: ah! so `@isOpen` is actually private api. 

I think perhaps I need to document the public/private ness of all the args 😅

NullVoxPopuli:
Thanks for doing this!!! <3
Sorry it took so long to get merged -- I'll get this released shortly

NullVoxPopuli:
Released in 0.12

aoifehannigan:
Thank you @NullVoxPopuli and @far-fetched!

Issue #130, opened 2022-02-18
Bump `@embroider/macros`


bertdeblock:
The version currently used by `ember-headlessui` causes a build error for me post install.

ember-cli: 4.1.1
node: 16.13.2
os: darwin arm64

bertdeblock:
This in turn also needs a bump of `ember-element-helper` it seems.

NullVoxPopuli:
it does -- but also, we should remove render-modifiers from this addon

bertdeblock:
What are the errors? Apologies if they are caused by my last PR.

NullVoxPopuli:
all `@embroider/*` deps need to be in-sync across a whole project.
So, the solution folks should use for the time being is:
`resolutions` (for yarn)
`overrides` (for npm 8+)

to force embroider deps to be all 1.2.0

PR #131, opened 2022-02-21
Upgrade ember-element-helper

cloke:
Addresses issue #130 which prevents this add-on from playing nicely with some Ember 4.0 apps.


bertdeblock:
I also included this change over in https://github.com/GavinJoyce/ember-headlessui/pull/135, so I think we can close this.

PR #136, opened 2022-03-28
Add Fallback Focus to Dialog

mansona:
Hey folks 👋 

We were having a few issues with Dialog sometimes failing because of the lack of a focusable element. It was failing with the following message from focus-trap: 

```
Your focus-trap must have at least one container with at least one tabbable node in it at all times
```

I looked into the other places in the codebase where focus-trap was being used and it does look like we're making use of the fallback focus functionality of `focus-trap` in the `Menu::Items` component: https://github.com/mansona/ember-headlessui/blob/dialog-fallback-focus/addon/components/menu/items.hbs#L15

I just lifted essentially the same implementation used for `Menu::Items` and applied it to Dialog 👍 I also wrote a test that verifies that the functionality actually works now 🎉 

Let me know if you have any questions.

### DIFF
--- a/addon/components/dialog.hbs
+++ b/addon/components/dialog.hbs
@@ -5,12 +5,14 @@
         id={{this.guid}}
         role='dialog'
         aria-modal='true'
+        tabindex='-1'
         ...attributes
         {{headlessui-focus-trap
           focusTrapOptions=(hash
             initialFocus=@initialFocus
             allowOutsideClick=this.allowOutsideClick
             setReturnFocus=this.setReturnFocus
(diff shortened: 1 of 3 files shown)


GavinJoyce:
👋 @mansona, thanks for this - I've come across the same issue, just didn't have time to address it. I'll cut a release soon

GavinJoyce:
https://github.com/GavinJoyce/ember-headlessui/releases/tag/v0.12.2

mansona:
woah that was fast! thanks @GavinJoyce 🎉

PR #139, opened 2022-04-06
fix rootElement when running in /tests

mansona:
We had disabled being able to use `http://localhost:4200/tests` in our app because a bunch of things were breaking and I tracked it down to the rootElement in `<Dialog />` not actually being correct. 

After a bit more investigation it turns out that the way that `setOwnConfig` was being used in config might have been the culprit: https://github.com/embroider-build/embroider/issues/1178

I've swapped that out for `ember-get-config` because I know it's working in this case and I have tested the dummy app + tests in this repo and our upstream app and it's now working 👍

### DIFF
--- a/addon/components/dialog.ts
+++ b/addon/components/dialog.ts
@@ -1,28 +1,14 @@
 import Component from '@glimmer/component';
+import { getOwner } from '@ember/application';
 import { action } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
 import { inject as service } from '@ember/service';
 import { typeOf } from '@ember/utils';
 
-import { getOwnConfig } from '@embroider/macros';
 import { Keys } from 'ember-headlessui/utils/keyboard';
 import { modifier } from 'ember-modifier';
 
 import type DialogStackProvider from 'ember-headlessui/services/dialog-stack-provider';
-
-/**
- * Expose the element that the `Dialog` should be "slotted" into
- *
- * This is exported _only_ for testing purposes; do not consider this API to be public
- *
- * @private
- */
-export function getPortalRoot() {
-  const { rootElement } = getOwnConfig();
-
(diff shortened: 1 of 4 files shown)


bertdeblock:
I feel that `ember-get-config` is overused, the app's config can be accessed in the component's constructor to determine the portal root.

mansona:
@bertdeblock yea sure if you'd prefer that I'm happy to switch it over 👍

bertdeblock:
It's up to the maintainers to make that decision of course, but I just wanted to weigh in on this.
Also because, `ember-get-config` explicitly mentions to only use it when you really don't have access to the container.

mansona:
@bertdeblock well I think it's also a preference thing. My previous attempt at this PR was conceptually closer to what was there before because it was making use for `@embroider/macros` (just with a bug). 

Also I'm a bit biased in favour of ember-get-config these days because I'm the maintainer of it 😉 

I am happy to set it up in the constructor if that gets this PR merged quicker 👍

mansona:
So I'm not exactly sure what is going wrong here 🤔  The only thing that I can think of is that there is some floating dependency in the ember-try runs that is causing the behaviour to change 

I did some investigation and it looks like when you click on the document body or the `#ember-testing` div (which is **under** the overlay) it will close all open dialogs instead of just the top one 🤔  The ember-get-config version of this had the same problem with tests too but I couldn't quite figure it out.

Any ideas?

mansona:
Any chance we could get this merged even though the test suite is blocked by an upstream error? I know it reduces our confidence in the suite but we know that this change doesn't cause the particular error at least 😞

GavinJoyce:
@mansona would you like commit / publishing access to `ember-headless`?

GavinJoyce (approved):

PR #140, opened 2022-04-06
add floating test from addon blueprint

mansona:
This is just taking the floating test from the Ember 4.x default addon blueprint: https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/.github/workflows/ci.yml#L33-L46

You can merge this if you like but it's mostly to prove a theory I have that https://github.com/GavinJoyce/ember-headlessui/pull/139 is not failing because of a change but instead because of some floating dependency 😞

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "webpack": "^5.64.1"
   },
   "resolutions": {
+    "@ember/test-helpers": "~2.6.0",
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
This is covered here; https://github.com/GavinJoyce/ember-headlessui/pull/144

PR #146, opened 2022-05-27
:bug: <listbox/> with transtion open on "enter" click

far-fetched:
### **What changed and why:**

Currently when using `<Listbox/>` component with composition with `<Transtion/>` it is not possible to use "Enter" key to open listbox.

Video from [demo](https://gavinjoyce.github.io/ember-headlessui/listbox/listbox-with-transition) (try yourself):  

https://user-images.githubusercontent.com/11621383/170680518-49a03b85-e6c0-43d9-a1dd-b2b39e429278.mp4

### **Fix:**

listbox with transition suffers from race condition. When “enter” click, [handler](https://github.com/GavinJoyce/ember-headlessui/blob/master/addon/components/listbox.js#L129) for `keyPress` is called – is sets flag `isOpen` to true. 
Then after a while, the [handler](https://github.com/GavinJoyce/ember-headlessui/blob/master/addon/components/listbox.js#L69) `onButtonClick` is also fired – this handler toggles `isOpen` property – so it reverts `isOpen` to false.
Everything above is executed on one “enter” click.

[link](https://github.com/GavinJoyce/ember-headlessui/runs/6623959069?check_suite_focus=true#step:7:185) to failed test.

### DIFF
--- a/tests/integration/components/listbox-test.js
+++ b/tests/integration/components/listbox-test.js
@@ -10,6 +10,8 @@ import { hbs } from 'ember-cli-htmlbars';
 import { module, skip, test, todo } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 
+import userEvent from '@testing-library/user-event';
+
 import {
   assertActiveElement,
   assertActiveListboxOption,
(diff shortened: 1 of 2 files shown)


tests/integration/components/listbox-test.js
```diff
@@ -869,6 +871,33 @@ module('Integration | Component | <Listbox>', function (hooks) {
...
+               as |options|>
+               <options.Option>
+                Option A
+               </options.Option>
+             </listbox.Options>
+            </Transition>
+         </Listbox>`);
+
+      getListboxButton()?.focus();
+      userEvent.keyboard('{Enter}');
```
far-fetched, line 893: Unfortunately, `triggerEvent` from ember-helpers does not simulate "Enter" like real browser. Only one handler is executed, thus this test would pass even without a fix for the relevant issue.

far-fetched: Yes, similar emulation we used with `dialog` test for [tab press](https://github.com/GavinJoyce/ember-headlessui/blob/master/tests/integration/components/dialog-test.js#L661).

NullVoxPopuli:
Thanks for doing this!!

PR #149, opened 2022-05-29
Remove Husky

NullVoxPopuli:
While working on this repo, I ran in to this error

```
.husky/pre-commit: 2: .: cannot open .husky/_/husky.sh: No such file
```

additionally, everyone should be setting:
```
npm config set ignore-scripts=true
yarn config set ignore-scripts true
```
because having any sort of post-install script active is a huge security risk -- and then disabling post-install scripts means husky won't work.

And since we want to ease effort-to-push, I think we should get rid of Husky.
If C.I. catching any issues, any maintainer can fix those.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-yarn lint-staged
--- a/package.json
+++ b/package.json
@@ -110,7 +110,6 @@
     "eslint-plugin-ember": "^10.5.5",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.3.1",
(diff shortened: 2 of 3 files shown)


GavinJoyce:
I'm not familiar with husky, do you know what we were using it for?

NullVoxPopuli:
it lints js, hbs, etc _before commit_, making it impossible to commit if there is a lint issue (unless you add `--no-verify` to your commit -- which I've had to do for every commit today because my local husky is broken, likely due to me having ignore-scripts)

bertdeblock:
You'll also want to remove the `.husky` folder and the `lint-staged` dependency. `husky` is only responsible for hooking into git. `lint-staged` is the one that executes the correct linters pre-commit for linting the staged files.

NullVoxPopuli:
@bertdeblock thanks! done in last commit

PR #150, opened 2022-05-29
Fix CI / don't use test.each

NullVoxPopuli:
the tests on the main branch are failing, so I removed `test.each` in favor of more native JS to try pin down which test specifically is causing issues.

that test will then be skipped / marked as todo.

whovever needs nested dialogs can work to unfix it :sweat_smile:


NullVoxPopuli:
Gonna merge this because all PRs are blocked until test stability is achieved. And while I still wouldn't consider tests stable, this is a step towards at least eliminating false negatives.

PR #116, opened 2021-12-07
Remove render modifiers

NullVoxPopuli:
wip

PR #154, opened 2022-06-02
Remove render modifiers

roomman:
This PR removes `@ember/render-modifiers`, in line with the discussion [here](https://github.com/GavinJoyce/ember-headlessui/pull/143#discussion_r886815312).

It replaces #116 but builds on the work done. I've only created this new branch because, after rebasing the original onto master, I found myself trying to fix unrelated issues. I trust that is okay with @NullVoxPopuli 🤞🏻

I've had to work around an issue, whereby we can't [yield local modifiers](https://github.com/ember-modifier/ember-modifier/issues/78#). I will circle back and do a bit of refactoring once that's resolved. For now, each sub-component has it's own, name spaced modifier, and references local actions that pass data up to their parent. If anyone knows of a better solution I'd love to hear it.

To pre-empt the question, I've had to make a significant change to the way that `<Menu>` keeps track of it's items. After implementing the local modifiers, I was seeing a ton of `You attempted to update 'items' on 'Menu', but it had already been used previously in the same computation...` errors. This makes sense, because `items` was a tracked property that was being consumed by several getters, and updated by actions in the same compute. To resolve this I've refactored `items` into a getter that returns an `_items` array, which is the thing that gets set. Again, if anyone has a better solution please shout. 👍🏻

Test are passing and I've had a good click around the demo app, so fingers crossed this is good to go.


NullVoxPopuli:
> we can't https://github.com/ember-modifier/ember-modifier/issues/78#.

can you expand on this?

I yield local modifiers just fine here: https://github.com/NullVoxPopuli/ember-popperjs/blob/main/ember-popperjs/src/components/popper-j-s.ts#L111

addon/components/dialog.hbs
```diff
@@ -17,8 +17,7 @@
         }}
         {{this.handleEscapeKey @isOpen this.onClose}}
         {{this.lockWindowScroll}}
-        {{did-insert (fn this.dialogStackProvider.push this)}}
-        {{will-destroy (fn this.dialogStackProvider.remove this)}}
+        {{this.stack}}
```
NullVoxPopuli, line 20: <3

addon/components/listbox/-button.js
```diff
@@ -0,0 +1,26 @@
...
+export default class ListboxButtonComponent extends Component {
+  registerButton = modifier((element) => {
+    this.registerButtonElement(element);
+
+    return () => {
+      this.unregisterButtonElement(element);
+    };
+  });
+
+  @action registerButtonElement(element) {
```
NullVoxPopuli, line 15: do these needs to be separate functions? this could probably be cleaned up a bit if inlined into the modifier?

addon/components/listbox/-label.js
```diff
@@ -0,0 +1,16 @@
...
+import { action } from '@ember/object';
+
+import { modifier } from 'ember-modifier';
+
+export default class ListboxLabelComponent extends Component {
+  registerLabel = modifier((element) => {
+    this.registerLabelElement(element);
+  });
+
+  @action registerLabelElement(element) {
```
NullVoxPopuli, line 11: same here -- why not put this function's content directly in the modifier?

addon/components/menu.js
```diff
@@ -104,25 +108,25 @@ export default class Menu extends Component {
 
   @action
   registerItem(item) {
```
NullVoxPopuli, line 110: I see what you mean about the double render error here. This code is kinda problematic (likely copied from the original implementation!)

Do you know if the list still updates of items are added post-render?

My concern is that we've last a bit of reactivity, and I wonder if it'd be better to instead detach from the tracking fram during render via an `await Promise.resolve()` in each of these methods.

Thoughts?

addon/components/transition/child.js
```diff
@@ -9,6 +11,28 @@ export default class TransitionChildComponent extends Component {
...
+    this.trackDomNode(element);
+
+    return () => {
+      this.untrackDomNode(element);
+    };
+  });
+
+  @action
+  trackDomNode(element) {
+    if (this.args.trackDomNode) {
```
NullVoxPopuli, line 25: stuff like this could be reduced to `this.args.trackDomNode?.(element)`
and then with the functions move into the modifier directly, we cut the code to 1/3rd :partying_face:

package.json
```diff
@@ -34,15 +34,14 @@
...
-    "@ember/render-modifiers": "^2.0.0",
     "@tailwindcss/typography": "^0.5.2",
     "ember-auto-import": "^2.2.0",
     "ember-cli-babel": "^7.26.3",
     "ember-cli-htmlbars": "^6.0.1",
     "ember-cli-typescript": "^5.1.0",
     "ember-concurrency": "^2.1.2",
     "ember-element-helper": "^0.6.0",
-    "ember-modifier": "^3.0.0",
+    "ember-modifier": "4.0.0-beta.1",
```
NullVoxPopuli, line 44: Not that this is a problem, but just calling it out,
this _may_ be a breaking change for consumers. _but_ since we are still pre v1, it's waay less of an issue

NullVoxPopuli (changes requested):
Thank you so much for doing this! this is great work! <3

Left a few comments / questions on some minor things

roomman:
> can you expand on this?

In the original PR, you had a `registerItemElement` modifier declared in `menu/menu-item` that yielded down to `menu/item-element` via a hash. That wasn't working as expected, and I attributed it to [this](https://github.com/ember-modifier/ember-modifier/issues/78#issuecomment-819053955) (I figured it was the hash causing the issue.) I've had a second run and it does work, so I will refactor. 

Not sure what happened there. 🤔

Nothing to see here. 🤣

addon/components/listbox/-button.js
roomman: No, they don't. I was experiencing some odd feedback from the test suite and found that breaking them out in this way resolved the errors. That said, like the yielding issue, those errors have gone and I've refactored as you've suggested.

Think I just had some bad luck with test feedback.

NullVoxPopuli: Yeah, i think this test suite can be a bit flaky at times

addon/components/menu.js
roomman: I won't pretend to understand the mechanics (feel free to explain when you have time) but what you proposed works and the refactoring is done. 👍🏻

addon/components/listbox/-label.js
```diff
@@ -0,0 +1,9 @@
+import Component from '@glimmer/component';
+
+import { modifier } from 'ember-modifier';
+
+export default class ListboxLabelComponent extends Component {
+  registerLabel = modifier((element) => {
+    this.args.registerLabelElement(element);
```
NullVoxPopuli, line 7: This can be done in a followup, but if `registerLabelElement`, itself, was a modifier, we don't need this class, and we can do `<div {{@registerLabelElement}}>`

addon/components/menu.js
```diff
@@ -103,20 +103,20 @@ export default class Menu extends Component {
   }
 
   @action
-  registerItem(item) {
+  async registerItem(item) {
```
NullVoxPopuli, line 106: long term, I think we need to find a different design for this -- this causes performance issues where we render _additional times_ for each item.

So, for 1000 items, we re-render 1000 times. not great. lol

(I assume Vue and React implementations have this same problem)

NullVoxPopuli (approved):
This is good for now! All the issues I have were pre-existing, and this PR is a huge improvement! (One less dependency!)

Thank you!

PR #168, opened 2022-10-24
[Issue-126] :: Implement Disclosure component 

Chris-Blesson:
In this PR, I have implemented the Disclosure component in ember similar to the react headless UI [component](https://headlessui.com/react/disclosure) 

I have added the relevant tests for the component functionality as well

I have also added the component to the docs

Please find the screen shots of the implementation
image
image

### DIFF
--- a/ember-headlessui/addon/components/disclosure.hbs
+++ b/ember-headlessui/addon/components/disclosure.hbs
@@ -1,9 +1,14 @@
-{{yield
-  (hash
-    isOpen=this.isOpen
-    Button=(component
-      'disclosure/button' toggleState=this.toggleState isOpen=this.isOpen
+<div
+  data-test-ember-headlessui-disclosure-wrapper
+  class='ember-headlessui-disclosure-wrapper'
+>
+  {{yield
+    (hash
+      isOpen=this.isOpen
+      Button=(component
+        'disclosure/button' toggleState=this.toggleState isOpen=this.isOpen
+      )
+      Panel=(component 'disclosure/panel' isOpen=this.isOpen)
     )
-    Panel=(component 'disclosure/panel' isOpen=this.isOpen)
-  )
-}}
+  }}
+</div>
--- a/ember-headlessui/addon/components/disclosure.hbs
+++ b/ember-headlessui/addon/components/disclosure.hbs
@@ -1,0 +1,9 @@
+{{yield
+  (hash
+    isOpen=this.isOpen
+    Button=(component
+      'disclosure/button' toggleState=this.toggleState isOpen=this.isOpen
+    )
+    Panel=(component 'disclosure/panel' isOpen=this.isOpen)
+  )
+}}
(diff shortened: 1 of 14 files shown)


Chris-Blesson:
Hi @NullVoxPopuli can you help me with the review for this PR

ember-headlessui/addon/components/disclosure.hbs
```diff
@@ -0,0 +1,14 @@
+<div
```
NullVoxPopuli, line 1: do we need the wrapping div?

ember-headlessui/addon/components/disclosure/button.js
```diff
@@ -0,0 +1,19 @@
+import Component from '@glimmer/component';
+
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { guidFor } from '@ember/object/internals';
+
+
+export default class Button extends Component {
+  guid = `${guidFor(this)}-tailwindui-disclosure-button`;
```
NullVoxPopuli, line 9: can we use the guid helper instead of having a backing class?

template-only components are more performant

test-app/app/components/disclosure/basic.hbs
```diff
@@ -0,0 +1,61 @@
+<div class='w-full px-4 py-4 pt-16'>
+  <div class='mx-auto w-full max-w-md bg-white shadow-md px-4 py-4 rounded-lg'>
+    <Disclosure as |disclosure|>
+      <span class='rounded-md shadow-sm'>
+        <disclosure.Button
+          @index={{1}}
```
NullVoxPopuli, line 6: we do we have index here? A consumer should never need to manage the index, yeah?

ember-headlessui/addon/components/disclosure/button.hbs
```diff
@@ -0,0 +1,12 @@
+<button
+  type='button'
+  aria-haspopup={{true}}
+  aria-controls={{if @isOpen this.itemsGuid}}
```
NullVoxPopuli, line 4: can we generate a single GUID in the parent component, just to reduce the number of repeat guid creations?

passing a guid (and then concatting suffixes) allows us to then also get rid of data-test-selectors (we don't want to ship these to consumers either, but that's a current problem all over the addon that we need to fix -- out of scope for this PR)

ember-headlessui/addon/components/disclosure/panel.js
```diff
@@ -0,0 +1,10 @@
+import Component from '@glimmer/component';
+
+import { tracked } from '@glimmer/tracking';
+import { guidFor } from '@ember/object/internals';
+
+export default class Panel extends Component {
+  panelGuid = `${guidFor(this)}-tailwindui-disclosure-panel`;
+  
+  @tracked isOpen = false;
```
NullVoxPopuli, line 9: why does this have its own isOpen state when the parent component also has isOpen?

NullVoxPopuli (changes requested):
Left some questions, overall very nice work!

PR #169, opened 2022-11-13
Tabs component

Shivareddy-Aluri:
implemented tabs component
Screenshot 2022-11-13 at 11 19 14 PM


ember-headlessui/addon/components/tabs-group.hbs
```diff
@@ -0,0 +1,5 @@
+<div data-test-ember-headlessui-tabs-wrapper ...attributes>
```
NullVoxPopuli, line 1: do we have to have this wrapping div?

test-app/app/templates/tabs/tabs-with-disable-state.hbs
```diff
@@ -0,0 +1,22 @@
...
+                    @disable={{if (eq index this.disabledToBeIndex) true}}
+                    class='w-full rounded-lg py-2.5 text-sm font-medium leading-5 ring-white ring-opacity-60 ring-offset-2 ring-offset-blue-400 focus:outline-none focus:ring-2 space-x-4 pointer-events-auto {{if (eq index this.activatedIndex) "bg-white shadow"}}'
+                    >
+                {{tabName}}
+                </tabs.Title>
+            {{/each}}
+        </div>
+        <div class="mx-auto w-full max-w-md shadow-md px-4 py-4 rounded-xl bg-white mt-4">
+            {{#each this.contents as |content index|}}
+                <tabs.Content class="content" @content-index={{index}}>
```
NullVoxPopuli, line 16: can we remove the index? and compare via object identity or some other way? using indexes makes looping inefficient / invalidate more frequently than we need it to

test-app/app/templates/tabs/tabs-with-disable-state.hbs
```diff
@@ -0,0 +1,22 @@
+<div class='w-full max-w-md px-2 py-16 mx-auto'>
+    <TabsGroup class="flex flex-col space-x-1 rounded-xl p-1" @onChange = {{this.onChange}} as |tabs|>
```
NullVoxPopuli, line 2: What are your thoughts on internalizing the onChange behavior by default?

can we abstract away the need for the consumer to configure this? (I think so!)

ember-headlessui/addon/components/tabs-group.js
```diff
@@ -0,0 +1,36 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+export default class TABSGROUP extends Component {
+  tabNames = [];
+  Contents = [];
+  @tracked currentTab = null;
+  @action
+  registerTabNames(e) {
+    this.tabNames = [...this.tabNames, e];
```
NullVoxPopuli, line 10: instead of using immutability patterns, can we switch to a `TrackedArray`? that will provide more optimal updates and not invalidate the whole collection

ember-headlessui/addon/components/tabs-group/-content.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="{{if this.renderContent '' 'hidden'}}"
```
NullVoxPopuli, line 1: I don't think we can assume a "hidden" class exists

ember-headlessui/addon/components/tabs-group/-content.js
```diff
@@ -0,0 +1,15 @@
...
+import { tracked } from '@glimmer/tracking';
+
+import { modifier } from 'ember-modifier';
+
+export default class Content extends Component {
+  @tracked element = null;
+  get renderContent() {
+    return this.element === this.args.selectedContent;
+  }
+  registerContent = modifier((e) => {
```
NullVoxPopuli, line 11: can we not have this side-effect here? I know this pattern is common in headlessui, but it's a problematic pattern, and we don't want to propagate it.

Derived data will be more performant, have fewer layout shifts, and will in general be _way_ easier to debug.

ember-headlessui/addon/components/tabs-group/-title.hbs
```diff
@@ -0,0 +1,11 @@
+<button
+  type='button'
+  data-test-headlessui-tabs-title='test-tabs-title-btn-{{@index}}'
+  id={{guid}}
+  disabled={{if @disable true false}}
+  {{on 'click' this.selectTab}}
+  {{this.registerTabs}}
```
NullVoxPopuli, line 7: remove this modifier please <3

NullVoxPopuli (changes requested):
Only got part way through so far, but some general notes:
 - don't have effects, instead, favor derived data patterns 
   - no registerItem/tab/whatever helpers/modifiers/etc
 - I think we can reduce the amount of glue code consumers need to use

PR #173, opened 2022-12-15
:bug: <Listbox> should scroll option into view only once

far-fetched:
Failing test reference: [click](https://github.com/GavinJoyce/ember-headlessui/actions/runs/3707544456/jobs/6284104779#step:7:726)

fixes: #137 

### **Issue:**

In this repo, we vastly used the pattern "register children elements" in [modifier](https://github.com/GavinJoyce/ember-headlessui/blob/master/ember-headlessui/addon/components/listbox.js#L174). Then when in modifier callback, we use `this.args.*` it will be reevaluated when that `args` changes <-- this is desired behavior.
But because of this mechanism, we have to be very careful what functions are called inside modifier. This bug is caused by calling `this.scrollIntoView` more than once. It affects the situation when user selects a new value (test) or new options are lazy loaded.

### **Solution:**

It turns out that we can scroll `listbox option` into viewport not from root component, but delegate it to children (`option`). There is an only one moment when option should be scrolled (when is selected – what is being checked in constructor of `option`). 

### **Before:**

Lazy loaded options with "jumpy" effect when promise resolved:

https://user-images.githubusercontent.com/11621383/208042239-f82d6d06-7fe1-4bbb-aeef-be64e91cb9cd.mp4

### **After:**

https://user-images.githubusercontent.com/11621383/208042199-cebf7586-679d-4130-9fbe-0475deb62da6.mp4

### DIFF
--- a/ember-headlessui/addon/components/listbox.hbs
+++ b/ember-headlessui/addon/components/listbox.hbs
@@ -23,6 +23,8 @@
       openListbox=this.openListbox
       closeListbox=this.closeListbox
       handleClickOutside=this.handleClickOutside
+      selectedValue=@value
+      scrollIntoView=this.scrollIntoView
     )
     Button=(component
       'listbox/-button'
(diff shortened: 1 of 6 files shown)


dmcnamara-eng:
Looks solid 👊  Tested manually locally by mutating list of options and working well ✅

PR #174, opened 2023-01-05
Fix closing animated modal

### LINKED ISSUE #172: `<Transition>` with `<Dialog>` breaks clicking the overlay to close
I noticed when perusing the examples today that the animation modal dialog example would not allow me to click outside of it to close the dialog. The non-animated one seems to work fine, and for the animated one if I remove the second `<t.Child>` it will also work. 

I wonder if we need to pass different parents to the transition or something? This is a complex z-index type of issue, since the overlay is nested in one parent and the modal content is nested in another and there is some fixed positioning.

I've been trying to debug this for awhile, but cannot seem to find a workaround. Perhaps we could use something like https://github.com/zeppelin/ember-click-outside and put that modifier on the modal dialog content, instead of handling the click on the overlay?

I'm happy to explore this as a solution if folks think it would be good!

rwwagner90:
This seems to fix being able to close the modal by clicking the overlay.

Closes #172

### DIFF
--- a/test-app/app/templates/dialog/dialog-modal-with-transition.hbs
+++ b/test-app/app/templates/dialog/dialog-modal-with-transition.hbs
@@ -39,7 +39,6 @@
           @leave='ease-in duration-200'
           @leaveFrom='opacity-100 scale-100'
           @leaveTo='opacity-0 scale-95'
-          class='transition-all transform'
         >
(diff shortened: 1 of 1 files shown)


far-fetched (approved):
:clap:

PR #175, opened 2023-01-26
FIX: add default type='button' to switch

rreckonerr:
Because, when used inside form the switch button reacts to clicks on Enter

### DIFF
--- a/ember-headlessui/addon/components/switch/button.hbs
+++ b/ember-headlessui/addon/components/switch/button.hbs
@@ -3,6 +3,7 @@
     id={{@guid}}
     role='switch'
     tabindex='0'
+    type='button'
     aria-checked='{{@isOn}}'
     aria-labelledby={{@labelGuid}}
     ...attributes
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
hello! thanks for your contribution?

for someone who's lost a bit of context around this (me), what problem is this change solving?

rreckonerr:
Hi @NullVoxPopuli ! Sorry for the poor PR description, my bad!

### Problem 
By default, the `type` attribute of `<button>` elements is `submit`. When `<Switch />` is used inside a `<form>` it's treated as a submit button, which is a confusing behavior.  In my case, hitting Enter inside a form triggered the change of the `<Switch />` input.

### Solution
We should provide default `type='button'` to the underlying `<Switch />` button element.

### Links
- [ember-template-lint/require-button-type](https://github.com/ember-template-lint/ember-template-lint/blob/0b4983a496b40452d3747a921c062cc5fa50c336/docs/rule/require-button-type.md)
- `headlessui-react/switch` implementation has the default `type='button'` attribute. Tests for this PR I took from their repo, see https://github.com/tailwindlabs/headlessui/blob/2f13496acb602479476e321471b051760d75e6e4/packages/%40headlessui-react/src/components/switch/switch.test.tsx#L63

NullVoxPopuli:
Thank you!!♥️

PR #178, opened 2023-03-06
Combobox refactor selected value

far-fetched:


test-app/tests/integration/components/combobox-test.js
```diff
@@ -215,7 +215,7 @@ module('Integration | Component | <Combobox>', function (hooks) {
         await render(hbs`
           <Combobox 
             @value={{this.value}} 
-            @onChange={{this.setValue}} 
+            @onChange={{this.onChange}}
```
far-fetched, line 218: In relevant test we specified `onChange` callback, not `setValue` (click expand up button)

far-fetched:
:wave: @balinterdi as you were the main contributor for `combobox` component, please have a look. Thanks in advance !

ember-headlessui/addon/components/combobox.js
```diff
@@ -21,15 +20,14 @@ const PREVENTED_KEYDOWN_EVENTS = new Set([
 ]);
 
 export default class ComboboxComponent extends Component {
-  @tracked selectedOptionGuids = new TrackedSet();
   @tracked _isOpen = this.args.isOpen || false;
   //TODO: `options` would be a better name as we store other stuff in it, too.
   @tracked optionElements = [];
   @tracked activateBehaviour = ACTIVATE_NONE;
   @tracked _activeOptionGuid;
+  options = [];
```
balinterdi, line 28: Also thinking of the `TODO` above, do we need both `optionElements` and `options`?

ember-headlessui/addon/components/combobox/-option.js
```diff
@@ -25,6 +36,7 @@ export default class ComboboxOptionComponent extends Component {
     if (this.args.disabled) return;
 
     this.args.setSelectedOption(this, e);
```
balinterdi, line 38: Is `this.args.setSelectedOption(this, e)` still needed? The component now handles sending the new selected values (or value) upwards, so I'm not sure this is still required.

balinterdi (commented):
I love how much of the code that keeps track of selected options has been removed from the top-level combobox component 👍🏼 ! 

If I see correctly, the `-option` components themselves send the new value for selected options. The only reason the register/unregisterOption methods are needed is so that the combobox can choose the first active/selected option, right?

Also, do we need both `this.registerOption` and `this.registerOptionElement` in the combobox component? Could we somehow merge them?

ember-headlessui/addon/components/combobox.js
far-fetched: As I described above, eventually I would like to remove modifier callback entirely (`this.registerOptionElement`) and `optionElements` collection as well. This PR is a first step.
Answering our question, if we can merge them (`registerOption` and `registerOptionElement`) — the ideal solution would be to remove register modifier, and only keep constructor / destructor if necessary (for select / active element) but not DOM element what we currently do.

ember-headlessui/addon/components/combobox/-option.js
far-fetched: Probably still necessary,

As inside this callback there is still some magic which relies on DOM elements. 

I think we can remove this as second step, when DOM elements (modifier) is no longer used.

balinterdi (approved):
Thank you for answering my questions: the changes look good to me, especially knowing that further PRs are coming.

dmcnamara-eng (commented):
Something is broken here with keyboard navigation

# Before:

https://user-images.githubusercontent.com/70960245/223684026-e6f60607-d9fe-4ed9-9a89-51b223b20014.mov

# After:

https://user-images.githubusercontent.com/70960245/223684039-2f3dc865-797d-4400-9422-878b9c256b2e.mov

far-fetched:
Thanks for testing @dmcnamara-eng. It should work right now.

dmcnamara-eng (approved):
> Thanks for testing @dmcnamara-eng. It should work right now.

WFM now 🙌

PR #182, opened 2023-09-01
Add preventDefault also for 'Enter' on menu item

johanrd:
Continuation of #80 and follows the implementation in https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/menu/menu.ts#L262


GavinJoyce:
Thanks for the PR @johanrd. Perhaps you might add some tests too?

PR #185, opened 2023-12-05
add types to Menu component

johanrd:


ember-headlessui/addon/components/menu.ts
```diff
@@ -1,19 +1,54 @@
...
 
 import { restartableTask, timeout } from 'ember-concurrency';
 
-export default class Menu extends Component {
+import type Button from './menu/button';
+import type Item from './menu/item';
+import type Items from './menu/items';
+
+interface MenuSignature {
+  Args: {
```
NullVoxPopuli, line 14: are all of these public API?
seems like many are private and shouldn't be exposed

johanrd: thanks, good catch, these types were actually meant to be on the button component. 

I have removed these now. If you find others, I would be happy to know.

ember-headlessui/addon/components/menu/item.ts
```diff
@@ -5,12 +5,26 @@ import { guidFor } from '@ember/object/internals';
...
+  };
+  Blocks: {
+    default: [{ isActive: boolean; isDisabled: boolean; Element: Element }];
+  };
+}
+
+export default class Item extends Component<MenuItemSignature> {
   guid = `${guidFor(this)}-tailwindui-menu-item`;
-  element;
+  element?: HTMLElement | null;
```
johanrd, line 24: TS complained about this one, probably since the element is registered from the item-element component, and not item directly?

PR #186, opened 2023-12-13
Draft of `<Popover>` component in typescript, inspired by the work in phorest/ember-headlessui

johanrd:
This is a draft for a typescript popover component, as an adaptation of the work by @achambers over at https://github.com/phorest/ember-headlessui/tree/feature/popover-component. 

@achambers I understand that this is not a good way make a PR, I apologize (I blame partly my git knowledge/headache). I consider this more like a typescript draft for inspiration/visibility. The component now works well in my monkey tests. Looking at the tests you have here (https://github.com/phorest/ember-headlessui/blob/feature/popover-component/tests/integration/components/popover-test.js), that could be a good basis to continue.


johanrd:
closing. At the moment, I would rather recommend to use the [native HTML popover attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)([polyfilled](https://github.com/oddbird/popover-polyfill)), together with e.g. [ember-velcro](https://github.com/CrowdStrike/ember-velcro):

```hbs
<button popovertarget={{@id}} popovertargetaction="toggle">Toggle</button>
<div id={{@id}} popover="auto" {{velcro (concat '[popovertarget=' @id ']')}}>Popover content</div>
```

PR #205, opened 2026-02-23
Add event to the dialog onClose action

barryofguilder:
This updates the `onClose` action for the `Dialog` component so that it passes the event with it. This will allow users to choose whether or not they want the dialog to close. A good example of this is if a user wants to prevent closing the dialog when clicking the overlay.

```handlebars
<button type="button" {{on "click" (set this "isOpen" true)}}>
  Trigger
</button>

<Dialog
  @isOpen={{this.isOpen}}
  @onClose={{this.handleClose}} as |d|
>
  <d.Overlay data-dialog-overlay />
  Contents
  <div tabindex="0"></div>
</Dialog>
```

```typescript
handleClose(event: Event) {
  let target = event.target as HTMLElement;

  if (target.hasAttribute('data-dialog-overlay')) return;
  this.isOpen = false;
}
```

---

### Original implementation

This adds a new optional argument to the `Dialog` component called `@closeOnOutsideClick` that will allow you to prevent closing the `Dialog` if you click the `Overlay` component or if you click outside the `Dialog`. This new argument defaults to `true` to preserve the existing behavior of always closing on outside click.

Here is an example using the new argument:

```handlebars
<button type="button" {{on "click" (set this "isOpen" true)}}>
  Trigger
</button>

<Dialog
  @isOpen={{this.isOpen}}
  @onClose={{set this "isOpen" false}}
  @closeOnOutsideClick={{false}} as |d|
>
  <d.Overlay />
  Contents
  <div tabindex="0"></div>
</Dialog>
```

### DIFF
--- a/ember-headlessui/addon/components/dialog.ts
+++ b/ember-headlessui/addon/components/dialog.ts
@@ -13,7 +13,7 @@ import type DialogStackProvider from 'ember-headlessui/services/dialog-stack-pro
 
 interface Args {
   isOpen: boolean;
-  onClose: () => void;
+  onClose: (event: Event) => void;
   as: string | typeof Component;
 }
 
@@ -35,7 +35,7 @@ export default class DialogComponent extends Component<Args> {
   });
 
   handleEscapeKey = modifier(
-    (_element, [isOpen, onClose]: [boolean, () => void]) => {
+    (_element, [isOpen, onClose]: [boolean, (event: Event) => void]) => {
       let handler = (event: KeyboardEvent) => {
         if (event.key !== Keys.Escape) return;
         if (!isOpen) return;
@@ -43,7 +43,7 @@ export default class DialogComponent extends Component<Args> {
         event.preventDefault();
         event.stopPropagation();
 
-        onClose();
+        onClose(event);
       };
 
       window.addEventListener('keyup', handler);
@@ -155,14 +155,14 @@ export default class DialogComponent extends Component<Args> {
       this.outsideClickedElement = target;
     }
 
-    this.onClose();
+    this.onClose(e);
 
     return true;
   }
 
   @action
-  onClose() {
+  onClose(event: Event) {
     if (this.dialogStackProvider.hasOpenChild(this)) return;
-    this.args.onClose();
+    this.args.onClose(event);
   }
 }
--- a/ember-headlessui/addon/components/dialog/-overlay.js
+++ b/ember-headlessui/addon/components/dialog/-overlay.js
@@ -26,6 +26,6 @@ export default class DialogOverlayComponent extends Component {
     event.preventDefault();
     event.stopPropagation();
 
-    onClose();
+    onClose(event);
   }
 }
--- a/test-app/tests/integration/components/dialog-test.js
+++ b/test-app/tests/integration/components/dialog-test.js
@@ -422,6 +422,29 @@ module('Integration | Component | <Dialog>', function (hooks) {
 
         assertDialog({ state: DialogState.InvisibleUnmounted });
       });
+
+      test('the `onClose` action sends the event with Escape', async function (assert) {
+        this.set('isOpen', false);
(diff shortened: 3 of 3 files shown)


barryofguilder:
@GavinJoyce Would love to see this one make it into the addon. We have use cases where we display a `Dialog` that has a form in it and we don't want people accidentally clicking outside the dialog to close it. We want them to either click the "Escape" key or to click one of the action buttons we have in the `Dialog`.

GavinJoyce:
Instead of adding a toggle, I wonder if we should add a `source` property to the `onClose` arguments? That would allow us to programatically control the behaviour:

```js
handleClose(source) {
    // source: 'outside-click' | 'escape'
    if (source === 'outside-click') return;
    this.isOpen = false;
  }
```

What do you think?

tniezurawski:
> Instead of adding a toggle, I wonder if we should add a `source` property to the `onClose` arguments? That would allow us to programatically control the behaviour:
> 
> ```js
> handleClose(source) {
>     // source: 'outside-click' | 'escape'
>     if (source === 'outside-click') return;
>     this.isOpen = false;
>   }
> ```
> 
> What do you think?

@GavinJoyce What about just passing the click `event` so the API consumer can decide based on the target what they want to do?

Boolean value, as proposed by @barryofguilder sounds ok, but also limiting. IMO, this should be a function.

In my app, I actually overrode the `allowOutsideClick()` function to be able to ignore some clicks. For example, when I have a dialog open, and an error notification (toast) is shown. Closing that toast notification shouldn't close the dialog, but it did - because it was an "outside click".

Having an event with a target can make it robust.

Snippet of how I'm dealing with it currently:

```js
import EmberHeadlessUiDialog from 'ember-headlessui/components/dialog.js';
import { action } from '@ember/object';

export default class extends EmberHeadlessUiDialog {
  @action
  allowOutsideClick(e) {
    // Existing code - start
    const target = e.target;

    if (target && target.tagName !== 'BODY') {
      this.outsideClickedElement = target;
    }
    // Existing code - end

    // Custom
    if (
      !this.#didClickOnNonClosingElement(target) &&
      !this.#didClickOnScrollbar(e)
    ) {
      this.onClose();
    }

    return true;
  }

// private methods where I check if our "[data-ui-notification]" was clicked for example
```

tniezurawski:
Forgot to write what I'd propose 😅 So I think, simply passing `event` to `this.onClose()` and let the API consumer decide what to do next would be enough:

```diff
@action
  allowOutsideClick(e: MouseEvent) {
    let target = e.target as HTMLElement;

    if (target && target.tagName !== 'BODY') {
      this.outsideClickedElement = target;
    }

+   this.onClose(e);
-   this.onClose();

    return true;
  }
```

Possibly with consuming the return of `this.onClose()` to return true/false from this action.

WDYT?

barryofguilder:
> Instead of adding a toggle, I wonder if we should add a `source` property to the `onClose` arguments? That would allow us to programatically control the behaviour:
> 
> ```js
> handleClose(source) {
>     // source: 'outside-click' | 'escape'
>     if (source === 'outside-click') return;
>     this.isOpen = false;
>   }
> ```
> 
> What do you think?

I think that would work.

GavinJoyce:
@tniezurawski 👍 I think passing the event in `onClose` makes most sense.

@barryofguilder would that work for you?

barryofguilder:
> @tniezurawski 👍 I think passing the event in `onClose` makes most sense.
> 
> @barryofguilder would that work for you?

That would work, it just adds a lot more complexity for the caller for the scenario that I need it for. I can update it to pass `event` though.

barryofguilder:
I've updated this to pass the event to the `onClose` action. As for tests, I wasn't sure the best way to do this, so I just added two: one to test the escape key and the other to test clicking the overlay.

GavinJoyce:
https://github.com/GavinJoyce/ember-headlessui/releases/tag/v0.16.1

thanks

PR #209, opened 2026-06-09
Fix importing-inject-from-ember-service deprecation

### LINKED ISSUE #208: Fix importing-inject-from-ember-service deprecation
Ref: https://deprecations.emberjs.com/id/importing-inject-from-ember-service/

We have one occurrence of `import { inject as service } from '@ember/service';` in the repo, and it's in the Dialog component.
Fixing the deprecation is a must to upgrade to Ember v7.

----
 
I proposed a PR with a fix.

tniezurawski:
Closes https://github.com/GavinJoyce/ember-headlessui/issues/208

There was only one occurrence of `inject as service` in the repo. 

From what I see [on the deprecation page](https://deprecations.emberjs.com/id/importing-inject-from-ember-service/), the `service` API is available from Ember v4.1. That raises the floor for the addon, which was currently tested against even v3.25. I moved it up to v4.4 because that was an LTS at the time.  This makes it a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,10 +124,8 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          - ember-3-25
-          - ember-lts-3-28
-          - ember-4-0
-          - ember-4-4
+          - ember-lts-4.4
+          - ember-lts-4.8
           - ember-release
           - ember-beta
           - ember-canary
--- a/test-app/package.json
+++ b/test-app/package.json
@@ -31,7 +31,7 @@
     "ember-headlessui": "*",
     "ember-modifier": "4.0.0-beta.1",
     "ember-page-title": "^7.0.0",
-    "ember-source": "~3.25.0",
+    "ember-source": "~4.4.0",
     "ember-test-selectors": "^6.0.0",
     "tailwindcss": "^3.0.24"
   },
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -124,10 +124,8 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          - ember-3-25
-          - ember-lts-3-28
-          - ember-4-0
-          - ember-4-4
```
tniezurawski, line 130: I'm not sure when this naming pattern has changed but I noticed here: https://github.com/ember-cli/ember-addon-output/blob/v4.12.3/.github/workflows/ci.yml#L60-L61 that when referring to version, the blueprint uses dot (`.`), like `ember-lts-4.4` instead of `ember-4-4`. 

I aligned it with the blueprint 🤷‍♂️ if that's ok.

test-app/package.json
```diff
@@ -31,7 +31,7 @@
     "ember-headlessui": "*",
     "ember-modifier": "4.0.0-beta.1",
     "ember-page-title": "^7.0.0",
-    "ember-source": "~3.25.0",
+    "ember-source": "~4.4.0",
```
tniezurawski, line 34: I might be wrong but the test application seems to be already running on v4.4 🤔 when it comes to all dependencies, with an exception for `ember-source` itself being set to v3.25.

tniezurawski:
@GavinJoyce could you review this? :)

dmcnamara-eng (approved):

dmcnamara-eng:
Thanks @tniezurawski  -- merging now.

